PR #2, opened 2017-08-09
PP-2481 Making tech-docs publishable in PaaS

kakumara:
This PR is to include tools in `tech-docs` so that its publishable to PaaS using a jenkins task.

Included `build.sh` and a `deploy.sh` to separate out the build/deploy bits.
Added a jenkins file so that build/deploy can be triggered from jenkins.
Using `cf` (cloudfoudry) push (similar to `pay-team-manual`) for pass deploy.


joshmyers:
Looks fine to me, good to know those scripts actually do what you want them to do. If it is copy/pasta from somewhere else...?

kakumara:
Thanks @joshmyers 
I've already tested this locally (although I had to fight with ruby versions,  rvm and middelman installations a bit). The only thing I assumed that these are installed in jenkins and no version conflicts.

And here is the published site https://govukpay-tech-docs.cloudapps.digital/#this-documentation

build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+BUNDLE="bundle"
+MIDDLEMAN="middleman"
+
+if ! which ${BUNDLE} > /dev/null; then \
+  echo -e "${red} ${BUNDLE} not installed, please install Bundler (http://bundler.io) ${NC}"
+  exit 1; \
+fi
+
+if ! which ${MIDDLEMAN} > /dev/null; then \
```
joshmyers: You probably don't want to check for a rubygem being present before you have actually installed it using bundler.

kakumara: ah! ok

kakumara:
@joshmyers addressed 👍

PR #22, opened 2017-08-21
Update to PCI DSS Section

jonathanglassman:
- Added content on PCI compliance requirements, based on number of transactions.
- Added table on SAQ A questionnaire


roryjhs:
This all looks pretty good but this is a section I think we'll need to test with partners and check with Bryan our QSA

PR #23, opened 2017-08-22
Update document title

36degrees:


source/index.html.md.erb
```diff
@@ -1,5 +1,5 @@
 ---
-title: GOV.UK Documentation Example
+title: GOV.UK Pay – Documentation
```
roryjhs: We're slowly standardising to have the unique title first followed by GOV.UK Pay. So the title should be:

Documentation – GOV.UK Pay

Can you change this or shall I?

36degrees:
Revised as requested!

PR #25, opened 2017-08-23
Update API explorer links to open in new tab

jonathanglassman:
Update API explorer links to open in new tab in following markdown files:
- Quick Start Guide
- Payment Overview
- API Reference
- Integration details
- Refunding payments


jonathanglassman:
@roryjhs I've amended the link text in line with guidance from Alistair Duggin (head of accessibility); each link now clearly states that it will open in a new window.

PR #32, opened 2017-09-05
Add in ePDQ screenshots

jonathanglassman:
Added in ePDQ screenshots to ePDQ section. A google doc with these screenshots has already been reviewed by Rory.

### DIFF
--- a/source/documentation/10-switching-to-production.md
+++ b/source/documentation/10-switching-to-production.md
@@ -36,6 +36,8 @@ To set up ePDQ to work with GOV.UK Pay, you must log into the ePDQ admin site an
 
 ### Add payment methods to your account
 
+![](images/epq-2.png)
+
 1. Log into the [ePDQ admin site](https://payments.epdq.co.uk/Ncol/Prod/BackOffice/login/index). On the homepage, go to _Configuration > Payment methods_ and select _Choose new payment methods_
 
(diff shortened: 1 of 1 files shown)


jonathanglassman:
Completely agree and I tried to make it consistent; however the numbered bullets would break for some reason in some cases depending on where I put the screenshots. I will investigate if I can today and let you know.

NeilBookless:
I don't know if my experience helps but...

I would tend to put the image first - I think the user will scan the image,
either consciously or subconsciously ('F' scan pattern), before reading the
written instruction.  [I think] this image pre-scan gives the user some
context for the written instruction and possibly some prior
understanding before they read the text, which requires a slightly
different cognitive set.

Best regards
Neil

-- 


*Neil Bookless *

Technical Author, GOV.UK Verify


GOV.UK Verify <https://identityassurance.blog.gov.uk/> | @GOVUKverify
<https://twitter.com/GOVUKverify>

The White Chapel Building | 10 Whitechapel High Street | London | E1 8QS
M: +44 (0) 7748 377 970

jonathanglassman:
Neil, I agree, I will try to figure out why putting the image before the ordered list sometimes breaks the formatting and make it consistent. Not sure if it's the tech doc tool or markdown.

PR #34, opened 2017-09-11
Make explicit that user-cancellation means "failed"

heathd:


source/documentation/06-api-reference.md
```diff
@@ -53,7 +53,7 @@ The response will include a ``status`` value as described in the table below, an
 | started      | User has visited next_url and is entering payment details                                      | false          |
 | submitted    | User has submitted payment details but has not yet clicked Confirm                             | false          |
 | success      | User successfully completed the payment                                                        | true           |
-| failed       | User didn't complete the payment, due to invalid or expired payment details, fraud block, etc. | true           |
+| failed       | User cancelled or otherwise didn't complete the payment, due to invalid or expired payment details, fraud block, etc. | true           |
```
alexbishop1: This phrasing makes it seem like invalid payment details can cause a cancellation which isn’t strictly true. Might be nice to shoehorn expiry in here too. Not quite sure how to phrase it. Maybe we need input from a tech writer.

bazbremner:
rebuild

PR #52, opened 2017-10-16
PCI DSS Tweaks

jonathanglassman:
Amended PCI DSS content in line with Bryan Scaife email from 12 October 2017


jonathanglassman:
Thanks Rory. I would keep referring to the external guidance for the SAQ A otherwise we'll have an additional maintenance overhead; happy to discuss if necessary.

PR #53, opened 2017-10-16
Update Switching to Live section

jonathanglassman:
Updated content in line with Till / Rory email from 12 October 2017: 
- Moved "Set up GOV.UK Pay account credentials" content to "Switching to Live" section so that it applies to both WorldPay and ePDQ rather than just ePDQ.
- Added link to quick start guide section on generating an API key.
- Added links to ePDQ setup section.

### DIFF
--- a/source/documentation/10-switching-to-production.md
+++ b/source/documentation/10-switching-to-production.md
@@ -46,10 +46,10 @@ If you want to use 3D Secure authentication for your payments, ask your Worldpay
 
 To set up ePDQ to work with GOV.UK Pay, you must log into the ePDQ admin site and:
 
-1. [Add payment methods to your account](https://govukpay-docs.cloudapps.digital/#add-payment-methods-to-your-account).
-1. [Set up account security parameters](https://govukpay-docs.cloudapps.digital/#set-up-account-security-parameters).
-1. [Set up notification settings](https://govukpay-docs.cloudapps.digital/#set-up-notification-settings).
-1. [Set up an API user](https://govukpay-docs.cloudapps.digital/#set-up-an-api-user).
+1. Add payment methods to your account.
+1. Set up account security parameters.
+1. Set up notification settings.
+1. Set up an API user.
 
 ### Add payment methods to your account
 
--- a/source/documentation/10-switching-to-production.md
+++ b/source/documentation/10-switching-to-production.md
@@ -2,9 +2,27 @@
 
 Once you have finished testing with your sandbox account, here are the steps you will need to take to switch over from testing to live.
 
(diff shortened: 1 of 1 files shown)


source/documentation/10-switching-to-production.md
```diff
@@ -2,9 +2,27 @@
...
+1. Go to _My Services_ and click on the _ePDQ Live_ service.
+1. Go to _Settings > Account Credentials > Edit credentials_.
+1. Complete the fields on this page:
+  - _PSP ID_ - enter your PSP ID for ePDQ
+  - _Username_ - enter the API user’s username
+  - _Password_ - enter the API user’s password
+  - _SHA-IN passphrase_ - enter the SHA-IN passphrase created on the Data and origin verification page
+  - _SHA-OUT passphrase_ - enter the SHA-OUT passphrase created on the Transaction feedback page
+1. Click _Save credentials_ to go back to the _Account Credentials_ page.
+
```
mrlumbu: this doesn't quite seem to be displaying correctly, at least as far as github markdown's parser is concerned
screen shot 2017-10-18 at 12 10 16

source/documentation/10-switching-to-production.md
```diff
@@ -28,11 +46,10 @@ If you want to use 3D Secure authentication for your payments, ask your Worldpay
...
 
-1. Add payment methods to your account.
-1. Set up account security parameters.
-1. Set up notification settings.
-1. Set up an API user.
-1. Set up GOV.UK account credentials.
+1. [Add payment methods to your account](https://govukpay-docs.cloudapps.digital/#add-payment-methods-to-your-account).
+1. [Set up account security parameters](https://govukpay-docs.cloudapps.digital/#set-up-account-security-parameters).
+1. [Set up notification settings](https://govukpay-docs.cloudapps.digital/#set-up-notification-settings).
+1. [Set up an API user](https://govukpay-docs.cloudapps.digital/#set-up-an-api-user).
```
mrlumbu: is there a reason why you list these links and then go through the content contained in those links in detail underneath. Feels a bit confusing and different from what we do for worldpay?

source/documentation/10-switching-to-production.md
jonathanglassman: It did the same thing when I first created the content, but then when it went live it was ok. I think it is the github parser.

source/documentation/10-switching-to-production.md
jonathanglassman: I added the links in so that users could jump to the section they want to refer to if necessary. However, happy to remove if you think it is confusing.

source/documentation/10-switching-to-production.md
mrlumbu: ok cool

source/documentation/10-switching-to-production.md
mrlumbu: I found it a bit confusing, but it might just be me. Maybe just the fact that it wasn't consistent with worldpay section

source/documentation/10-switching-to-production.md
jonathanglassman: I've removed the internal links for now. Are you happy to approve this PR?

PR #56, opened 2017-11-02
Switching to live updates

jonathanglassman:
Updates to the Switching to Live section:
- Changes to the GOV.UK Pay account setup to remove ePDQ-only information
- Specify that SHA-IN and SHA-OUT passphrases are plain text strings

Related to story: https://payments-platform.atlassian.net/browse/PP-2831

### DIFF
--- a/source/documentation/10-switching-to-production.md
+++ b/source/documentation/10-switching-to-production.md
@@ -89,7 +89,7 @@ Go to _Configuration > Technical information_
 1. On the _Technical information_ page, click the _Data and origin verification_ tab.
 1. Scroll to the _Checks for e-Commerce & Alias Gateway_ section.
 1. Leave the _URL of the merchant page containing the payment form that will call the page: orderstandard.asp_ field blank.
-1. Enter a strong plaintext string SHA-IN passphrase; this will be used when setting up the GOV.UK account credentials.
+1. Enter a strong SHA-IN passphrase (plain text, not a hash); this will be used when setting up the GOV.UK Pay account credentials.
 1. Scroll down to the _Checks for Barclaycard Direct Link_ section.
 1. Leave the _IP address_ blank.
 1. Enter the same SHA-IN passphrase as the _Checks for e-Commerce & Alias Gateway_ section and click _Save_.
@@ -118,7 +118,7 @@ Go to _Configuration > Technical information_
 ![](images/epq-6.png)
 
 1. Scroll to the _All transaction submission modes > Security for request parameters_ section.
-1. Enter a strong plaintext string SHA-OUT passphrase; this will be used when setting up the GOV.UK account credentials.
+1. Enter a strong SHA-OUT passphrase (plain text, not a hash); this will be used when setting up the GOV.UK Pay account credentials.
 1. Leave the _Basic Authentication Credentials_ blank.
 1. Set _Timing of the request_ to _For each offline status change (payment, cancellation, etc.)_
 1. Enter **https://notifications.payments.service.gov.uk/v1/api/notifications/epdq** into the _URL on which the merchant wishes to receive a deferred HTTP request, should the status of a transaction change offline_ field.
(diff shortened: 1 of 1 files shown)


source/documentation/10-switching-to-production.md
```diff
@@ -91,7 +89,7 @@ Go to _Configuration > Technical information_
 1. On the _Technical information_ page, click the _Data and origin verification_ tab.
 1. Scroll to the _Checks for e-Commerce & Alias Gateway_ section.
 1. Leave the _URL of the merchant page containing the payment form that will call the page: orderstandard.asp_ field blank.
-Enter a strong SHA-IN passphrase; this will be used when setting up the GOV.UK account credentials.
+1. Enter a strong plain text string SHA-IN passphrase; this will be used when setting up the GOV.UK account credentials.
```
daniele-occhipinti: Looks good. Minor thing: is it `plain text string`, or `plain-text string`? You maybe write `plain text`, but if you use it as an apposition for `string`, you may need the hyphen?

jonathanglassman: @daniele-occhipinti good point. I've checked and it seems to be written as "plaintext" when it is used for human-readable data that is viewable without any sort of decryption required. I'll amend.

@roryjhs @alexbishop1 any further comments on this?

source/documentation/10-switching-to-production.md
```diff
@@ -91,7 +89,7 @@ Go to _Configuration > Technical information_
 1. On the _Technical information_ page, click the _Data and origin verification_ tab.
 1. Scroll to the _Checks for e-Commerce & Alias Gateway_ section.
 1. Leave the _URL of the merchant page containing the payment form that will call the page: orderstandard.asp_ field blank.
-Enter a strong SHA-IN passphrase; this will be used when setting up the GOV.UK account credentials.
+1. Enter a strong plaintext string SHA-IN passphrase; this will be used when setting up the GOV.UK account credentials.
```
alexbishop1: Not too sure about this: I’ve only ever seen “plaintext” (with no spaces) used in contrast to “ciphertext” (i.e. unencrypted versus encrypted text); not sure it’s appropriate in this context.

I’m also a bit concerned that the extra words are taking “strong” too far away from “passphrase” (it almost seems like it’s a modifier for “string”, which we may be able to drop).

And I’ve just noticed that it says “GOV.UK account credentials” rather than “GOV.UK Pay account credentials”, which we may as well fix while we’re here.

Could we go with something like this?

“Enter a strong _SHA-IN passphrase_ (plain text, not a hash); this will be used when setting up the GOV.UK Pay account credentials.”

jonathanglassman: Agree, I've made the changes

jonathanglassman:
@daniele-occhipinti @alexbishop1 are you both happy with the changes?

daniele-occhipinti:
I have approved the changes. Hopefully @alexbishop1 is happy as well. Thanks @jonathanglassman

jonathanglassman:
@alexbishop1 I've added in Worldpay specific information, please let me know if you have any comments on this

source/documentation/10-switching-to-production.md
```diff
@@ -10,15 +10,26 @@ To set up your live account credentials:
...
+- _Password_ - enter the API user’s password
+- _SHA-IN passphrase_ - this passphrase is created on the [_Data and origin verification_ page](/#set-up-checks-for-e-commerce-amp-alias-gateway)
+- _SHA-OUT passphrase_ - this passphrase is created on the [_All transaction submission modes_ page](/#set-up-security-for-request-parameters)
+- click _Save credentials_ to go back to the _Account Credentials_ page
+
+### Worldpay
+
+- _Merchant ID_ - enter your merchant ID for Worldpay
+- _Username_ - enter the API user's username
+- _Password_ - enter the API User's password
```
alexbishop1: I think for Worldpay it’s referred to as the XML username and password

jonathanglassman:
@alexbishop1 I've changed the worldpay content to XML username and XML password

PR #79, opened 2018-02-09
API error codes update

jonathanglassman:
- Added section on API errors caused by payment status
- Added subtitle to distinguish API error caused by API call from errors driven by payment status

### DIFF
--- a/source/documentation/06-api-reference.md
+++ b/source/documentation/06-api-reference.md
@@ -85,7 +85,9 @@ These are the known status codes you are likely to receive:
 
 ## API error codes
 
-When an error occurs, you will receive these API codes in the body of the response.
+### API errors caused by an API call
+
+When your API call results in an error, you will receive these API codes in the body of the response.
 
 This is the format of the general JSON error response body:
 
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
 | General | P0920 | Request blocked by security rules | Our firewall blocked your request. See Troubleshooting section for details. |
 | General | P0999 | GOV.UK Pay is unavailable | The GOV.UK Pay service is temporarily down. |
 
+### API errors caused by payment statuses
+
+The API error codes in this section are driven by payment status.
+
+| Error code | Meaning | Cause |
+|:---|:---|:---|
+| P0050 | Payment provider returned an error | Multiple possible causes, for example a configuration problem with the payment provider, or incorrect login credentials; contact us, quoting the error code |
+| P0040 | Payment was cancelled by service | The service cancelled the payment |
+| P0010 | Payment method rejected | Multiple possible causes, for example the user does not have enough money in account, or 3D Secure authentication failure |
+| P0020 | Payment expired | A payment has not been confirmed and completed within one hour of being created; if the payment was already authorised, GOV.UK PAY will send a cancellation to the payment provider |
+| P0030 | Payment cancelled by user | User clicks on the "Cancel payment" button during the payment journey; if the payment was already authorised, GOV.UK PAY will send a cancellation to the payment provider |
+
+You can see which transactions have these error codes using either of the following methods:
+
+- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a failed transaction
+- use the API to look at the `state` object in a transaction's API response, for example:
+
+    ```
+    {
+        "amount": 2000,
+        "state": {
+            "status": "failed",
+            "finished": true,
+            "message": "Payment expired",
+            "code": "P0020"
+        },
+    }
+    ```
 
 ## Card types
 
--- a/source/documentation/06-api-reference.md
+++ b/source/documentation/06-api-reference.md
@@ -140,7 +140,7 @@ These error codes provide more information about why a request failed.
 
 The API error codes in this section are driven by payment status. You can see which payments have these error codes using either of the following methods:
 
-- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a  payment with a "failed", "cancelled" or "error" status.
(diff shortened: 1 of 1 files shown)


source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
 | General | P0920 | Request blocked by security rules | Our firewall blocked your request. See Troubleshooting section for details. |
 | General | P0999 | GOV.UK Pay is unavailable | The GOV.UK Pay service is temporarily down. |
 
+### API errors caused by payment statuses
+
+The API error codes in this section are driven by payment status.
+
+| Error code | Meaning | Cause |
+|:---|:---|:---|
+| P0050 | Payment provider returned an error | Multiple possible causes, for example a configuration problem with the payment provider, or incorrect login credentials; contact us, quoting the error code |
```
alexbishop1: Is there any particular reason the error codes aren’t listed in numerical order?

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
...
+| P0040 | Payment was cancelled by service | The service cancelled the payment |
```
alexbishop1: Maybe “Your service” would be clearer?

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
...
+| P0010 | Payment method rejected | Multiple possible causes, for example the user does not have enough money in account, or 3D Secure authentication failure |
```
alexbishop1: I wonder if this could be reworded to emphasise that it’s fundamentally caused by a problem relating to the payment information the user entered. Something like, “The user’s payment was rejected because [various reasons]…”. Fraud checks are probably another reason worth highlighting here.

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
...
+| P0020 | Payment expired | A payment has not been confirmed and completed within one hour of being created; if the payment was already authorised, GOV.UK PAY will send a cancellation to the payment provider |
```
alexbishop1: “The payment was not confirmed…” might reflect the finality of the situation better

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
...
+| P0030 | Payment cancelled by user | User clicks on the "Cancel payment" button during the payment journey; if the payment was already authorised, GOV.UK PAY will send a cancellation to the payment provider |
```
alexbishop1: Past tense (“User clicked”) would be more consistent the other explanations

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
...
+
+You can see which transactions have these error codes using either of the following methods:
+
+- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a failed transaction
+- use the API to look at the `state` object in a transaction's API response, for example:
```
alexbishop1: I wonder if this part should be moved up so that it’s before the table (in the same way as the _API errors caused by an API call_ section). The self-service stuff could remain after the table, as this is an API-focused section of the docs.

Also, I think we always talk about a “payment” rather than a “transaction” in the context of the API.

source/documentation/06-api-reference.md
jonathanglassman: No reason, have changed

source/documentation/06-api-reference.md
jonathanglassman: Changed

source/documentation/06-api-reference.md
jonathanglassman: Changed

source/documentation/06-api-reference.md
jonathanglassman: Changed

source/documentation/06-api-reference.md
jonathanglassman: Changed

source/documentation/06-api-reference.md
```diff
@@ -134,6 +136,34 @@ These error codes provide more information about why a request failed.
 | General | P0920 | Request blocked by security rules | Our firewall blocked your request. See Troubleshooting section for details. |
 | General | P0999 | GOV.UK Pay is unavailable | The GOV.UK Pay service is temporarily down. |
 
+### API errors caused by payment statuses
+
+The API error codes in this section are driven by payment status. You can see which payments have these error codes using either of the following methods:
+
+- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a failed payment
```
alexbishop1: (This is actually unchanged from the first commit but I only just noticed it)

We have “failed” as one of our external payment states (alongside “created”, “started”, “submitted”, “success”, “cancelled” and “error”). The error codes listed here could belong to “failed” payments but also “cancelled” and “error” payments. So perhaps we should say “click on an unsuccessful payment” or something.

source/documentation/06-api-reference.md
```diff
@@ -140,7 +140,7 @@ These error codes provide more information about why a request failed.
 
 The API error codes in this section are driven by payment status. You can see which payments have these error codes using either of the following methods:
 
-- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a failed payment
+- log into the [GOV.UK Pay admin site](https://selfservice.payments.service.gov.uk/), go to _Transactions_ and click on a  payment with a "failed", "cancelled" or "error" status.
```
alexbishop1: Really picky: can you use “curly quotes” here?

Also, there appears to be two spaces between “a” and “payment”.

alexbishop1: As discussed, it seems that magic in the build process converts "straight quotes" into “curly quotes”, so we’re OK there

PR #88, opened 2018-04-04
Upgrade to latest tech-docs-template

tijmenb:
This PR upgrades the tech docs to the latest version, which uses the [govuk_tech_docs gem](https://github.com/alphagov/tech-docs-gem) to pull in CSS, JS, fonts, images and layouts. This means that a lot of files can be removed and only customised code and configuration remains. You can read more about it in https://github.com/alphagov/tech-docs-template/pull/144.

I'm upgrading projects that haven't upgraded yet as part of the GOV.UK firebreak, in the hope that if we are all on the same version it's more likely that sites will keep up to date and contribute back to the main project.

## Automatic upgrades

Since future upgrades will be done by updating the `govuk_tech_docs` dependency, we can set up [Dependabot](https://dependabot.com) to do the upgrades for us. Whenever a new version of the gem is available, Dependabot will raise a PR to update the version. It looks something like this: https://github.com/alphagov/govuk-developer-docs/pull/858.

To enable this, ask a GitHub owner to add the repo to [repos Dependabot can access](https://github.com/organizations/alphagov/settings/installations/87197). After that, you can go to the [dependabot dashboard and click "Add project"](https://app.dependabot.com/accounts/alphagov/repos) to turn on automatic updating.

## How to review

There should be no changes in the layout or behaviour after this upgrade. I've checked this by comparing https://govukpay-docs.cloudapps.digital  to the new site.

---

Part of the Q4 GOV.UK 🔥 firebreak: https://trello.com/c/QbspvRc2


heathd:
Thank you @tijmenb 🎉

PR #90, opened 2018-04-10
Self service payments

jonathanglassman:
Added content on self-service payments to the following sections:
- Before you start
- Quick start guide
- Payment flow overview
- Switching to live

### DIFF
--- a/source/documentation/04-quick-start-guide.md
+++ b/source/documentation/04-quick-start-guide.md
@@ -78,3 +78,49 @@ The ``return_url`` is the URL of a page on your service that the user will be re
 1. Go to the [service admin site](https://selfservice.payments.service.gov.uk/). Select **Transactions** at left. You’ll see the payment you just made.
 
 ![](images/transaction+list+image+4.png)
+
+## Self-service payments
+
+You may want to use self-service payment links rather than integrate with the GOV.UK Pay API if your service:
+
+- is non-digital and uses paper applications to process payments
+- is non-digital and has low transaction volumes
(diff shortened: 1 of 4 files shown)


source/documentation/04-quick-start-guide.md
```diff
@@ -78,3 +78,49 @@ The ``return_url`` is the URL of a page on your service that the user will be re
...
+1. Sign in to your [GOV.UK Pay account](https://selfservice.payments.service.gov.uk/).
+1. Select [Create a payment link](https://selfservice.payments.service.gov.uk/create-payment-link) in the Dashboard.
+1. Select __Create a payment link__.
+
+### Set payment link information
+
+On the __Set payment link information__ page:
+
+1. Complete the __Title__ field.
+1. Edit the __URL__ field if required. The __URL__ field follows this format: __https://<i></i>gov.uk/payments/SERVICE_NAME/TITLE__. For example, _https://<i></i>gov.uk/payments/dvla/renew-your-licence_. You can edit the __TITLE__ section of the URL. Refer to the [GOV.UK guidance on URL standards](https://www.gov.uk/guidance/content-design/url-standards-for-gov-uk).
```
soniaturcotte: This is called in web address (url) in the admin tool
Most users can’t edit the field unless the specific url already exists, which is pretty unlikely. So this part might confuse the majority of users.

jonathanglassman: We have used URL in the tech docs rather than web address because URL is in the tech writing style guide. We should keep it internally consistent.

PR #95, opened 2018-05-15
Support #2788803 Updating documentation

belindac:
All documentation links are pointing to 1.0.1 which has been deprecated by v1.0.2.
There is also no documentation specifying the expected values for the different payment type states when searching for a payment when using the api.
The payment states can be found when viewing https://govukpay-docs.cloudapps.digital/#payment-status-lifecycle, but should also be linked to in the example column for state in the gelato documentation.
The card type values can be found when viewing https://docs.payments.service.gov.uk/#card-types, but should also be linked to in the example column for card_brand in the gelato documentation.

solo @belindac

### DIFF
--- a/source/documentation/06-api-reference.md
+++ b/source/documentation/06-api-reference.md
@@ -17,10 +17,10 @@ The same base URL is now used for testing and live. The API key you use determin
 
 For full details of each API action, see the API Browser:
 
-- <a href="https://gds-payments.gelato.io/docs/versions/1.0.1/resources/general" target="blank">General functions</a> (link opens in new window)
-- <a href="https://gds-payments.gelato.io/docs/versions/1.0.1/resources/payment-id" target="blank">Payment ID functions</a> (link opens in new window)
+- <a href="https://gds-payments.gelato.io/docs/versions/1.0.2/resources/general" target="blank">General functions</a> (link opens in new window)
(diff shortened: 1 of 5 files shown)


source/documentation/06-api-reference.md
```diff
@@ -43,19 +43,19 @@ This diagram gives an overview of the payment status lifecycle and the possible
...
-| started      | User has visited next_url and is entering payment details                                      | false          |
-| submitted    | User has submitted payment details but has not yet clicked Confirm                             | false          |
-| success      | User successfully completed the payment                                                        | true           |
-| failed       | User attempted to make a payment but the payment did not complete. Possible reasons: user cancelled; user entered payment details but never confirmed; payment declined due to insufficent funds; payment declined due to fraud or other checks | true           |
-| cancelled    | Your service cancelled the payment using an API call or the self-service site.                 | true           |
-| error        | Something went wrong with GOV.UK Pay or the underlying Payment Service Provider.               | true           |
+| API value | status value | Meaning                                                                                        | finished value |
```
heathd: https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/validation/PaymentSearchValidator.java#L24

bahmady:
Thanks @belindac 

Unfortunately I contacted Gelato support and there isn't a way to link to a <latest>, so as far as the docs are concerned we need to change versions manually as they become deprecated. I'll try to think of something non-manual to solve this problem.

PR #107, opened 2018-06-29
Change versioning section info to reflect latest

bahmady:
Also:

* removes callout for version of API 
* adds 1.0.1 and 1.0.2 to revision history


heathd:
looks good

we should probably make a statement that we adhere to [semver](https://semver.org/) as well.

PR #111, opened 2018-07-09
Add line about adhering to SemVer

bahmady:
Per subject line, spun out of @heathd's comment on https://github.com/alphagov/pay-tech-docs/pull/107#issuecomment-402148047


heathd:
let's leave this for now. I'd like to write something about versioning at get feedback from users about it before putting it in the documentation.

PR #112, opened 2018-07-09
update docs for financial reporting integration

katiebates:
because content is about 18 months out of date...


bahmady:
Sorry for hijacking but are any of the other similar lines also no longer relevant? e.g.

> This will soon be fixed as part of the beta.

> Soon, we hope to let you check which version of our API you’re currently running by checking your dashboard.

> This will soon be fixed as part of the beta.

(not sure how best to point to them in this PR other than to just copy the language for copy/paste into a CTRL+F)

PR #114, opened 2018-07-11
Change “Edit organisation details” to “Organisation details”

alexbishop1:
This matches the words that are actually on the page


bahmady:
I've only just gotten access to a Pay account so only just been able to check this :D

PR #115, opened 2018-07-18
Change language in intro re. "government"

bahmady:
Removes emphasis on government and makes it clear it's for all public sector


bahmady:
Can't request review from @katiebates for some reason but can tag her. Not sure what's going on

PR #131, opened 2018-08-10
Simplify support information

bahmady:
Proposal to simplify the information in the 'support' section. Much of this is duplicated unnecessarily from what's sent to partners when they sign on, some is incorrect or outdated, and some doesn't conform to our house style.

### DIFF
--- a/source/documentation/16-support-contact-and-more-information.md
+++ b/source/documentation/16-support-contact-and-more-information.md
@@ -1,28 +1,8 @@
-# Support, contact and more information
+# Support
 
-## Support hours and service
+## Incident severity 
 
-In hours:
-
-- On site support: Mon-Fri 9.30am-5.30pm
-- Full support of GOV.UK Pay for all ticket priorities
-- 2 dedicated support people, with additional cover from the whole team as required
-
-Out of hours:
-
-- On call support: Mon-Fri 5.30pm-9.30am Sat-Sun 24x7
-- P1 priority only
-- 2 dedicated on call support people, with escalation as necessary
-
-## Asking for support
-
-You can ask for support via email. The severity of the incident affects who you should ask for support:
-
-- For all non-P1 support queries, get in touch by emailing [govuk-pay-support@digital.cabinet-office.gov.uk](mailto:govuk-pay-support@digital.cabinet-office.gov.uk). This will raise a support ticket.
-
-- In emergencies, Service teams can raise a P1 incident by emailing us. This will raise a high priority ticket and __could alert the team in the middle of the night, so please only use it for P1 incidents__. Please do not share this address with people outside your team.
-
-Incident severity is defined below:
+GOV.UK Pay incident severities are defined as follows:
 
 | Classification | AKA | Example | Initial response time | Update Time |
 | :--- | :--- | :--- | :--- | :--- |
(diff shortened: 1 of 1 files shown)


bahmady:
Happy to put this in the test environment for review if it's easier but if you look in 'Files changed' and then 'View' it should show you the content more or less as it will be rendered live.

bahmady:
Also something which occurs to me is that no mention is made of Bank Holidays?

source/documentation/16-support-contact-and-more-information.md
```diff
@@ -31,77 +14,31 @@ Incident severity is defined below:
 | P3 | Significant | Users experiencing intermittent or degraded service due to platform issue | 1 business day or next working day | 2 business days |
 | P4 | Minor | Component failure that is not immediately service impacting | 2 business days | weekly |
```
bahmady: @katiebates @heathd  can we add something here along the lines of "normal BAU functions" e.g. switching an account to live

bahmady:
Should we add something along the lines of P4 being the priority level for a BAU request, e.g. switching an account to live?

katiebates:
The simplification is a helpful change and I agree it's time to delete the sections about newsletters, press etc (or move them out of docs, to another product page). 

A few specific comments: 
We often tell partners that support is handled by the team, as a way to show that we have a direct feedback loop into product improvement, and the devs have a vested interest in ensuring the platform is operating robustly. Would it be useful/ appropriate to mention it here? It was alluded to in the last set of content: (e.g. "At all times, we have a dedicated Support Lead and Secondary (rotated weekly) who draw on additional support as required from other members of the team.. Additionally, a manager will act as the Support escalation route.")

In the Contact Us section, there is a reference to "support requests" but then to "support queries" - i don't think it's completely clear that a support request and query are the same thing. Also, I think it could be clearer that support requests come in via email to us, but we'll reply using the Zendesk ticketing system.

bahmady:
> We often tell partners that support is handled by the team, as a way to show that we have a direct feedback loop into product improvement, and the devs have a vested interest in ensuring the platform is operating robustly. Would it be useful/ appropriate to mention it here? It was alluded to in the last set of content: (e.g. "At all times, we have a dedicated Support Lead and Secondary (rotated weekly) who draw on additional support as required from other members of the team.. Additionally, a manager will act as the Support escalation route.")

Here I will just entirely defer to you; if you would prefer that this be in this section that's your decision as far as I'm concerned. My opinion would be that it's not vital, simply because I'm accustomed to seeing users make use of sections like this to find out where and how to find support and when they can, and nothing much else. 

> In the Contact Us section, there is a reference to "support requests" but then to "support queries" - i don't think it's completely clear that a support request and query are the same thing. Also, I think it could be clearer that support requests come in via email to us, but we'll reply using the Zendesk ticketing system.

On the first point, agreed, this was a holdover that I should have caught. On the second point, I've tried to make it slightly clearer - I'm not sure if this level is what you were suggesting, but I can happily make even clearer still if not.

katiebates:
I'm hesitant about spelling out switching to live specifically as a P4, as the SLA for P4 is 2 days. We can (and do) generate live accounts very quickly now, and want to emphasise how quickly we can get services taking payments.

bahmady:
@katiebates I agree with you; I think that makes sense. Based on what else we discussed about this content, are you happy with it as-is or do you think it needs any further changes?

PR #132, opened 2018-08-13
Style/language changes in 'Contribute' page

bahmady:
- 'MIT License' is a proper noun and should be treated in US English
- The MIT License link shouldn't go to a Wikipedia page: in this case we shouldn't actually need to link to a generic page and it would be far better if we could link to our actual licence 
- 'Open source' should not be capitalised and should be hyphenated when used as compound adjective
- We shouldn't use "email us" as a link; it should be spelled out fully per GOV.UK style 
- The header 'Open-source components on GOV.UK Pay' gets the point across better
- Other minor language changes


alexbishop1:
Each of our repositories has its own copy of the MIT License, so there’s not something canonical we can link to. https://opensource.org/licenses/MIT is probably the best we have if we want to link somewhere authoritative.

PR #133, opened 2018-08-13
Radical changes to 'Versioning'

bahmady:
I'm not necessarily proposing these changes as a "please approve this or make suggestions then approve it". I'd rather use this PR as a place to gather feedback and have a discussion about the points I'd like to raise and if they're sensible, _then_ we can approve them. 

I don't believe we should have a 'versioning' section that needs manual updating every time a new version is introduced. For the same reason I don't believe we should have a revision history table (which is in any case empty at the moment).

Making statements like "In the future we will..." is difficult on live documentation, too, if we're not sure of the timeframe. 

I'd also like to know if these points are true:

> We’ll also send you an email to let you know about any new versions we release. 
> We’ll support each version of the early beta GOV.UK Pay API for at least 1 month after we issue an upgrade notice. As the API matures, we may increase this support period to 3 or 6 months.


katiebates:
I agree it's good to simplify this section - it's evident that we haven't kept it up to date (e.g. revision history) and it removes another dependency for the team. 

The API explorer allows users to explore differences between previous versions if they want to. That should remove the need to have a revision history or an API changelog (that Stripe provides, for example - https://stripe.com/docs/upgrades#api-changelog). So as long as whatever replaces Gelato allows users to explore/ understand differences with previous API versions, I'm ok with that. 
But then.. relying on the API explorer doesn't allow users to see exactly when we've added new fields to the API (as this wouldn't create a new version)... is it useful to systematically note this kind of change? Since we'll be adding new parts to the API with the card parity features. Or should we rely on ensuring the relevant parts of the documentation are updated to reflect new additions? (though they won't be marked as new additions..) 

This line has been deleted: "When we add new properties to the JSON responses, the Pay API version number will not change. You should develop your service to ignore properties it does not understand." - this sounds pretty standard, but it might be worth including it as it could be a common mistake? (we've made it recently ourselves on products). Or it could go in a 'best practices' section?

Sending an email for a new version: we would do this, but we don't have a systematic way of doing it (i.e. we don't have a standard mailing list for all services). But Till is looking at how we can better update services for new changes. Which might be an email or something else.

bahmady:
> relying on the API explorer doesn't allow users to see exactly when we've added new fields to the API (as this wouldn't create a new version)... is it useful to systematically note this kind of change? Since we'll be adding new parts to the API with the card parity features. Or should we rely on ensuring the relevant parts of the documentation are updated to reflect new additions? (though they won't be marked as new additions..)

This really depends on the approach we want to take, on which I would defer. As a matter of course, we still need to consider the overall implementation and how we present the API to users. We would need to explore ways in which we could automatically note changes that constitute new versions or at least factor that into the consideration of the platform(s) we settle on in our presentation. Generally speaking, I would advise against relying on individuals to update documentation to manually make a change to indicate a version change. That being said, it would probably be easy enough to write a simple script though to pull down the latest version in whichever explorer we're using (assuming we are using one) and have that in this documentation. I  imagine I could do that myself with our Gelato implementation, straightforwardly. 

> This line has been deleted: "When we add new properties to the JSON responses, the Pay API version number will not change. You should develop your service to ignore properties it does not understand." - this sounds pretty standard, but it might be worth including it as it could be a common mistake? (we've made it recently ourselves on products). Or it could go in a 'best practices' section?

Happy to add this back and will do so in a commit shortly after finishing this comment. Just to reiterate, this PR was more of a call to action inspired by the fact that I believe it can be better. 

> Sending an email for a new version: we would do this, but we don't have a systematic way of doing it (i.e. we don't have a standard mailing list for all services). But Till is looking at how we can better update services for new changes. Which might be an email or something else.

OK, thanks for answering this. I'll focus on documenting things as they are now.

bahmady:
I've added the content back in you pointed to, @katiebates; I've reverted the filename to 'versioning.md' as a result though and changed the title back, since I think my suggested change to "Updating to the latest API" would no longer make sense.

bahmady:
@katiebates merge conflicts finally fixed, sorry 😞 

@heathd, Katie is happy with this content but as agreed this needs your approval, as well, so please review at your earliest convenience. 

As ever we can always build on this iteratively in future; I just want the existing documentation to express current the state of things as much as possible.

PR #136, opened 2018-08-15
Independent self-service section

bahmady:
### Context
The self-service section is substantive enough that it should be in its own section. Two independent sources of user feedback have suggested this, as well, including one explicitly. 

Changes to the directory structure as a whole to enable the creation of an independent self-service section

### Guidance to review
Preview this at https://new-self-service-section.cloudapps.digital/#about-gov-uk-pay


katiebates:
Definitely agree that this should have its own section. 
But i wonder if it should be called "self-service payment links" or "payment links" - our product page around it calls them "self service payments" but then refers to them immediately as "payment links". I don't know the precise background to the naming decision though - probably worth checking in with Mark Woods over it. 

I know the proposed changes aren't about the actual content in the section, but adding my twopence: 
1. the docs could link to the product page - https://www.payments.service.gov.uk/self-service-payments/
2. not sure if it's helpful for the documentation to explain the step by step instructions for setting up a payment link?

bahmady:
> But i wonder if it should be called "self-service payment links" or "payment links" - our product page around it calls them "self service payments" but then refers to them immediately as "payment links". I don't know the precise background to the naming decision though - probably worth checking in with Mark Woods over it.

I'll catch up with @mpwoods on this. 

> the docs could link to the product page - https://www.payments.service.gov.uk/self-service-payments/

This is a straightforward change to make but ...

> not sure if it's helpful for the documentation to explain the step by step instructions for setting up a payment link?

... this is less straightforward. Happy to do this but would need to go back to the drawing board and ask for developer pair writing time.

bahmady:
Actually, I said adding the product page was straightforward but I think it's less so. This is a good example where I need to be aligned with @mpwoods and Karl... We can certainly link to the product page but should we really be trying to sell the links here if we do have that page? 

Perhaps a bigger task than I'd anticipated.

PR #139, opened 2018-08-21
Change link to Wikipedia page re. PSPs to one from FCA

bahmady:
### Context
The current landing page links to a Wikipedia article on PSPs; if possible, we should try to avoid using Wikipedia links since they can be easily edited by anybody, in theory 

Changes Wikipedia link to an FCA one: https://www.handbook.fca.org.uk/handbook/glossary/G2619.html

### DIFF
--- a/source/documentation/02-about-govuk-pay.md
+++ b/source/documentation/02-about-govuk-pay.md
@@ -4,7 +4,7 @@ GOV.UK Pay is currently in beta development. The payment pages are responsive
 and work on both desktop and mobile.
 
 The platform can connect your service to different [payment service providers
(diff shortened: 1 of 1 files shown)


katiebates:
I agree we should use FCA rather than Wikipedia. The wiki article gives a fuller explanation of what a PSP is, but (on balance) we should link to a more reputable source.

bahmady:
The other option is we could write our own content on what a PSP is @katiebates ...

bahmady:
Alternative PR https://github.com/alphagov/pay-tech-docs/pull/141

bahmady:
Closing this—we're going to go with a variation of #139

PR #141, opened 2018-08-24
Remove link to Wikipedia

bahmady:
### Context
The existing content describes PSPs using a Wikipedia link, which is also un-cited. 

This is an alternative to https://github.com/alphagov/pay-tech-docs/pull/139

I wrote some content about PSPs. This would serve as an alternative to https://github.com/alphagov/pay-tech-docs/pull/139

See also: ongoing discussions re. whether we should carry on using a glossary


bahmady:
cc @karlchillmaid @mpwoods

PR #144, opened 2018-08-28
Fix support table not rendering

bahmady:
### Context
After the recent support changes, an old bug with a table not rendering properly came up.

A workaround/fix to this. Also the status page isn't rendering as a linked URL so I've embedded it in a link. 

### Guidance to review
As of right now: https://docs.payments.service.gov.uk/support_contact_and_more_information/#incident-severity


bahmady:
Yeah it's not ideal :( but it does work for some reason. I've tested locally.

PR #149, opened 2018-08-28
Removes glossary

bahmady:
### Context
Removes the glossary. I've been through the links in the rest of the documentation and haven't found any links to this, so this should be sufficient. I've also built this locally to test, and nothing breaks. 

Removes the files in the docs that render as the glossary.


heathd:
If you think this isn't helping people's comprehension of the docs, then fine. 

It's still accessible from the left nav, isn't it? So people might find it via the nav and read it. 

I agree it most of the terms seem to be fairly self evident, but I'm not a good example of the audience.

bahmady:
I've had a few discussions surrounding this and I'm fine to leave it in, but I think the main argument against removing it is that it would need someone to maintain it if we were to expand it. And as you say, the terms at the moment aren't hugely useful. 

Without GA it's difficult to determine if people are actually using this right now, but even if they are, as it stands I would suggest it won't be improving people's comprehension of the content. I'd like to see if during the course of improving the Pay documentation as a whole, we are able to manage without a glossary in terms of the design of the content; if we find that user research suggests people are struggling or even better that they suggest a glossary would be useful, we can reintroduce the glossary (assuming we merge this PR). 

This PR should automatically remove it from the left-hand nav, as well: the tool picks that data up from the `index` files.

PR #152, opened 2018-09-06
Switching to live overhaul

bahmady:
### Context
The "Switching to live" section needed a lot of work and updating. 

Changes and corrections to the "Switching to live" section.

### Guidance to review
I've pushed these changes to the test environment at https://govukpay-docs-test.cloudapps.digital/switching_to_live/ so reviewers can see it how it will appear _in situ_.

I will leave this PR open for at least a couple of days for the GOV.UK Pay team to have the opportunity to review it and let me know of any corrections and clarifications etc. I should also note that any and all feedback is welcome. @heathd and @katiebates have already reviewed separately, as have @karlchillmaid and @mpwoods at different points. 

Please note that the GDS tech docs tool still has some limitations despite recent significant improvements: for large pieces of content like this it is still not perfect, but very few tech docs tools are. 

Please also note that even if further clarifications etc. are needed after this is approved/published etc. it will be easy to iteratively do so.


bahmady:
@heathd I don't see why this isn't building via CI 🤔

source/documentation/set_up_a_live_worldpay_account.md
```diff
@@ -0,0 +1,164 @@
...
+
+__SENT_FOR_REFUND__
+
+__REFUNDED__
+
+__REFUNDED_BY_MERCHANT__
+
+__REFUSED__
+
+__REFUSED_BY_BANK__
```
bahmady: @katiebates a small thing but in P-377 there is a discrepancy in the image between "TEST" and "PRODUCTION": would you mind verifying?

bahmady:
Aside from my comment here FAO @katiebates https://github.com/alphagov/pay-tech-docs/pull/152/files#r215992298 I'm happy with this now.

katiebates:
1. "When you sign up for GOV.UK Pay, the team will contact you with specific contact details." This refers to the emergency contact details that Pay gives, but it's not super clear that's what it's about. Maybe something like "When you sign up for GOV.UK Pay, the team will give you emergency contact details" would be a little clearer?

2. Do we want to add a note about checking card types enabled, checking email settings (for emails to be on/off, and there'll be future options), billing address settings (when we add them)? When we talk through with a service the switch to live process, we often talk them through at least card types. But also it should also be pretty-self explanatory in self-service, and has no dependencies on other suppliers that a service would need to know about (which is the case for some of the 3DS configurations).. so maybe it's fine without?

3. In the "Request a live GOV.UK Pay account" section, we could also ask services to also supply us with the service name and what type of PSP account. We handle this at the moment by basically being in touch with the service, but it would be a bit simpler to have them send it to us. This is the info we currently need when setting up a live account: https://pay-team-manual.cloudapps.digital/support/creating-gateway-accounts/#creating-a-gateway-account 

4. "Test your configuration" section for Smartpay + ePDQ describes testing using payment links; "How to check Worldpay settings" is the new section we wrote, which describes using API or payment links.

bahmady:
@katiebates 

1. and 3. No problem, will change shortly 

2. We could perhaps add this in latterly: might be easier to review in a different pull request?

4. Ah this is my oversight, Worldpay has diverged a little. I'll change Smartpay and ePDQ to match Worldpay sections shortly

bahmady:
@katiebates I've added information which I think covers your points 1, 3, and 4. You should see this now at https://govukpay-docs-test.cloudapps.digital/switching_to_live/#switching-to-live

Point 2 I think may be more work: perhaps we could pick that up after the 24th?

PR #153, opened 2018-09-06
PP-4107 Add some documentation about Welsh payment pages

alexbishop1:
First attempt to add some documentation about our shiny, new Welsh language support for card payment pages.

Not even sure this is in the right place but it was the most appropriate location I could find.

We may want more detailed instructions on how to add a Welsh service name (it’s kind of hidden: log in, go to _Switch service_ and choose _Edit name_ next to your service).

@bahmady — could you take a look and finesse this into something usable please?

### DIFF
--- a/source/documentation/integration-details.md
+++ b/source/documentation/integration-details.md
@@ -60,6 +60,17 @@ If a user does not have enough funds in their account to make a payment, the cur
 You can cancel a payment that is not yet in a final state by using the <a href="https://gds-payments.gelato.io/api-explorer/gov-uk-pay-api/versions/1.0.2/payment-id/cancel-payment" target="blank">Cancel payment</a> API call (link opens in new window).
 
 
+## Welsh language support
+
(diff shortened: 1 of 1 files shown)


bahmady:
Thanks for this @alexbishop1: do you mind if I copy this and put it into a new section called "Optional features" on the same hierarchy as the other h1s in the documentation? I was thinking of putting Welsh in there alongside instructions on custom branding.

alexbishop1:
Yeah, that sounds great. I only put it in _Integration details_ because I couldn’t find anywhere better. We’ll be adding a few new optional features (delayed capture, corporate surcharging etc.) in the future, so it will be nice for them to have a home.

bahmady:
@alexbishop1 are you happy if I close this given we now have https://github.com/alphagov/pay-tech-docs/pull/154? Thanks again for your help.

alexbishop1:
Yeah, we can close this one.

PR #160, opened 2018-09-24
First attempt at drafting some content on not needing HTTPS for sandbox

bahmady:
### Context
@tillwirth asked me to specify in the documentation that it's possible to use HTTP with sandbox accounts when making payments.

First attempt at restructuring some parts of the documentation to reflect the above request, mindful that we still enforce HTTPS for live accounts.

### Guidance to review
Needs review from the product side and/or tech side to verify the language makes sense.

### DIFF
--- a/source/documentation/security.md
+++ b/source/documentation/security.md
@@ -82,4 +82,6 @@ Your service manager may also be asked to supply extra evidence on your internal
 
 GOV.UK Pay follows [government HTTPS security guidelines](https://www.gov.uk/service-manual/domain-names/https.html). The Hypertext Transfer Protocol Secure (HTTPS), which involves the Transport Layer Security  (TLS) protocol is used by the platform to authenticate servers / clients and to provide secure connections.
 
-Your government service will only be able to integrate with the GOV.UK Pay if it also uses HTTPS.
+To take a payment with a live GOV.UK Pay account, you must use HTTPs. 
+
+To take a payment with a sandbox account, [you can use HTTP](/testing_govuk_pay/#testing-gov-uk-pay).
--- a/source/documentation/security.md
+++ b/source/documentation/security.md
@@ -82,6 +82,8 @@ Your service manager may also be asked to supply extra evidence on your internal
 
 GOV.UK Pay follows [government HTTPS security guidelines](https://www.gov.uk/service-manual/domain-names/https.html). The Hypertext Transfer Protocol Secure (HTTPS), which involves the Transport Layer Security  (TLS) protocol is used by the platform to authenticate servers / clients and to provide secure connections.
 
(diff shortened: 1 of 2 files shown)


source/documentation/security.md
```diff
@@ -82,4 +82,6 @@ Your service manager may also be asked to supply extra evidence on your internal
 
 GOV.UK Pay follows [government HTTPS security guidelines](https://www.gov.uk/service-manual/domain-names/https.html). The Hypertext Transfer Protocol Secure (HTTPS), which involves the Transport Layer Security  (TLS) protocol is used by the platform to authenticate servers / clients and to provide secure connections.
 
-Your government service will only be able to integrate with the GOV.UK Pay if it also uses HTTPS.
+To take a payment with a live GOV.UK Pay account, you must use HTTPs. 
+
+To take a payment with a sandbox account, [you can use HTTP](/testing_govuk_pay/#testing-gov-uk-pay).
```
alexbishop1: I think this could benefit some context to say it really relates to the return URL. If that’s too specific, perhaps it could say that a service that integrates with Pay has to use HTTPS but there’s an exception for sandbox accounts to make development easier.

source/documentation/testing-govuk-pay.md
```diff
@@ -9,6 +9,11 @@ When testing, you’ll need to ensure:
  - REST calls succeed with 200 API codes
  - you’ve tested the user journey from your service to the payments platform using end-to-end/smoke tests
 
+You can make payments using HTTP with sandbox accounts, when testing.
```
alexbishop1: It’s only the service that can be HTTP; the rest (interactions with Pay via the API and web browser) has to be HTTPS

source/documentation/security.md
bahmady: I'm fine with either but not quite sure how to phrase it if we go with the first. Do you think the first would be more useful to the user?

alexbishop1: In terms of HTTPS, there’s two things really.

Firstly, all direct communication between the service and Pay must be HTTPS. This is implied by the API URLs and usually not onerous to do, so it doesn’t require much calling out.

The second aspect is that the return URLs must be HTTPS (except for sandbox). This is something we know can be a pain and the requirement has surprised people in the past (which is why we added the sandbox exemption).

If we just say you have to use HTTPS, I think a lot of people will only think about the communication between the service and Pay and not realise we mean the return URLs as well (except the sandbox exemption).

bahmady:
Thanks @alexbishop1, I've tried to implement your suggestions whilst keeping the language simple: if you have the time to review the new version and let me know if it's better/makes sense or is too simple/I've lost the meaning, would be very grateful. Thanks again 👍

bahmady:
Makes sense! Thanks for the explanation 👍 , have edited.

PR #164, opened 2018-09-26
Quick start guide changes

bahmady:
### Context
As part of my ongoing review of the documentation I am going through each section in turn and making suggested changes. Much of this will be simple but some of the changes in given PRs will need more thorough reviews by subject matter experts (SMEs). This is because I may introduce language which could be phrased more clearly by an SME, and of course I may make mistakes.

The quick start guide did not confirm to GDS style entirely and I made some changes to the language to make it more active and straightforward for users to follow. I also copied some information we have from elsewhere about testing with return URLs in a sandbox environment (i.e. not needing HTTPS), in case users only get this far into their journey before wondering about that.

### DIFF
--- a/source/documentation/quick-start-guide.md
+++ b/source/documentation/quick-start-guide.md
@@ -1,19 +1,25 @@
 # Quick start guide
 
-GOV.UK Pay is based on REST principles with endpoints returning data in JSON format, and standard HTTP error response codes. The platform API allows you to:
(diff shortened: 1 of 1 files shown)


bahmady:
@katiebates hmm I don't know how to make those nice diagrams like Sonia used to make. Shall I enquire with Stephen before publishing this or publish this as-is first?

katiebates:
i think it's fine to publish this now - it's only a minor change to the page.

PR #165, opened 2018-09-29
Reorganise folder structure

heathd:
- introduce numbered folder structure
- inline markdown text in the index files of sub-pages

no textual changes, just moved stuff around

PR #166, opened 2018-10-02
Simplify directory structure

bahmady:
### Context
The directory structure is unnecessarily complicated at the moment, splitting off a separate `documentation` directory which contains mostly Markdown files which could simply be in Index files 

Move Markdown content into `index.html.md.erb` files and remove `documentation` directory altogether.

### Guidance to review
There should be no changes to the end user experience (I have tested this locally) with these changes, despite how extensive they look. I am only doing this to make developers and other contributors to the documentation's lives easier. 

This is an alternative to https://github.com/alphagov/pay-tech-docs/pull/165

https://github.com/alphagov/pay-tech-docs/pull/165 keeps the documentation directory intact and introduces numbering to Markdown files. In the long term, I agree numbering the Markdown files would be useful, once we are in a more "final" state; however, as of right now we are still making regular and continuous changes and reorganising the information in the documentation as a whole, so my opinion is that numbering the Markdown files now would make less sense. 

With that said, I'm fine with either PR and their proposals: but there's no need for both PRs.


bahmady:
For the time being we've decided to adopt this PR to stop merge conflicts with further changes but we can revert these changes if necessary and adopt #165, including the commit about Markdown numbering

PR #171, opened 2018-10-08
Fix language (request/response)

bahmady:
### Context
The current content erroneously references an HTTP response header as a a request.

An attempt to fix ☝️ , though it may be slightly clunky and need checking/improving/correcting by an expert.

### DIFF
--- a/source/payment_flow_overview/index.html.md.erb
+++ b/source/payment_flow_overview/index.html.md.erb
@@ -64,9 +64,7 @@ For example:
 + **description**: a human-readable description of the payment. This will be shown to the end user on the payment pages and to your staff on the GOV.UK Pay admin site (maximum 255 characters, and must not contain URLs)
 + **return_url**: an HTTPS URL on your site that the user will be sent back to once they have completed their payment attempt on GOV.UK Pay (this should not be JSON-encoded as backslashes will not be accepted)
 
-The **Create new payment** API call that the service will receive will be of the form:
-
-Example request:
+In the example, the **Create new payment** API call the service receives would give the following response header:
 
 ```response
 HTTP/1.1 201 Created
@@ -74,7 +72,7 @@ Content-Type: application/json
 Location: https://publicapi.pymnt.uk/v1/payments/icus7b4umg4b4g5fat4831es5f
 ```
 
-Example response:
+And the following response body: 
 
 ```javascript
 {


source/payment_flow_overview/index.html.md.erb
```diff
@@ -64,17 +64,15 @@ For example:
 + **description**: a human-readable description of the payment. This will be shown to the end user on the payment pages and to your staff on the GOV.UK Pay admin site (maximum 255 characters, and must not contain URLs)
 + **return_url**: an HTTPS URL on your site that the user will be sent back to once they have completed their payment attempt on GOV.UK Pay (this should not be JSON-encoded as backslashes will not be accepted)
 
-The **Create new payment** API call that the service will receive will be of the form:
-
-Example request:
+In the example, the **Create new payment** API call the service receives would give the following response header:
```
alexbishop1: The service isn’t receiving the **Create new payment** API call, it’s receiving the _response_ to the API call. I think something like this would do it:

“In the example, the response to the **Create new payment** API call would have headers like the following:”

(Using “like” because it’s not exact — replaying the exact example would give a different ID in the `Location` header each time)

source/payment_flow_overview/index.html.md.erb
```diff
@@ -64,17 +64,15 @@ For example:
...
 
 ```response
 HTTP/1.1 201 Created
 Content-Type: application/json
 Location: https://publicapi.pymnt.uk/v1/payments/icus7b4umg4b4g5fat4831es5f
```
alexbishop1: I know this was already like this but `publicapi.pymnt.uk` is our test environment, which services never see. It should be `publicapi.payments.service.gov.uk`.

source/payment_flow_overview/index.html.md.erb
```diff
@@ -64,17 +64,15 @@ For example:
...
 ```
 
-Example response:
+And the following response body:
```
alexbishop1: Maybe “The response body would look similar to the following:” or something? (Again, it’s not going to be exactly the same each time and we’re not going to update this example every time we add new properties etc.)

source/payment_flow_overview/index.html.md.erb
bahmady: Makes sense, thanks 👍 : do we really need the response header or just the body?

alexbishop1: The fact that it’s a 201 status code and that the `Location` header is a self-link to the payment is quite useful to call out.

bahmady:
Thanks @alexbishop1, I've made an attempt, let me know what you think. 

Also re. the point about `publicapi.pymnt.uk`; I think there may have been some other instances that needed changing in the same section of the docs, would you be able to look at where I've changed these and let me know if these changes were sensible? 

Many thanks 👍

source/payment_flow_overview/index.html.md.erb
```diff
@@ -91,11 +90,11 @@ Example response:
   "created_date": "2016-06-24T11:46:11.414Z",
   "_links": {
     "self": {
-      "href": "https://publicapi.pymnt.uk/v1/payments/icus7b4umg4b4g5fat4831es5f",
+      "href": "https://publicapi.payments.service.gov.uk/v1/payments/icus7b4umg4b4g5fat4831es5f",
       "method": "GET"
     },
     "next_url": {
-      "href": "https://www-integration-2.pymnt.uk/secure/bb0a272c-8eaf-468d-b3xf-ae5e000d2231",
+      "href": "https://www-integration-2.payments.service.gov.uk/secure/bb0a272c-8eaf-468d-b3xf-ae5e000d2231",
```
alexbishop1: `www-integration-2` should be `publicapi`

PR #172, opened 2018-10-09
New landing page for documentation

bahmady:
### Context
A new landing page for the technical documentation (i.e. where users end up when they first see it, unless they're following a specific link). Instead of the "About GOV.UK Pay" page, which I highly doubt anybody reads, they are now linked to the more appropriate (and more regularly maintained) features page on GOV.UK and in this page they are now given a simple description about the platform and what it does in case they end up on the docs first as part of their user journey - this is relatively unlikely but worth accounting for. The landing page also now offers an explanation of how to navigate through the technical documentation. 

Almost entirely new landing page 

### Guidance to review
Use the rich text preview to see how this would appear in situ

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -3,16 +3,21 @@ title: Documentation – GOV.UK Pay
 weight: 10
 ---
 
-# GOV.UK Pay technical documentation 
+# GOV.UK Pay documentation 
 
-You can use this technical documentation as a reference if you are a user of
-the GOV.UK Pay platform. 
+If you are a user of the GOV.UK Pay platform, you can use this documentation
+as a reference. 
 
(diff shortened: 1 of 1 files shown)


bahmady:
@katiebates I've made the changes after the in-person conversation with Till. You can see these at:

https://govukpay-docs-test.cloudapps.digital

bahmady:
Thanks @jonheslop. Apart from my typo in a commit ("effect" should be "affect" 😱) I think this is good to go, and yesterday @katiebates was happy with it after we incorporated the feedback from @tillwirth, who was happy to leave it to us. 

I'll go ahead and merge this shortly if nobody else has any feedback: we can always revert / add more information if needed.

PR #175, opened 2018-10-15
Simplify landing page

bahmady:
### Context
In hindsight I think this is better: this way we're not necessarily asserting that people should use the products and features page, but recommending it for people who are completely new to the platform.

Simplifies landing page a bit.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -5,15 +5,10 @@ weight: 10
 
 # GOV.UK Pay documentation 
 
-If you are a user of the GOV.UK Pay platform, you can use this documentation
-as a reference. 
-
-Use the left-hand navigation bar to move to different sections of the
+Use the left-hand navigation bar to learn about the different sections of the
 documentation.
 
-## New to GOV.UK Pay 
-
-If you are new to GOV.UK Pay, you can read about the platform on our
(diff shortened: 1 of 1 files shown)


katiebates:
I like having fewer titles on that page, and so the simplification is good, although i'm not convinced about asserting the product/ features page is for 'completely' new users, since (for example) you might be somewhat familiar to Pay, but never read the features page, and not read the optional features section, so not know about welsh language pages. also the features page has been relaunched, and the product page is likely to get an overhaul soon too. 

Re payment links - not sure about this one. There's now content about it on the features page, and it's on the product page too. But it's a great feature and under-used - but not sure that docs is the right place to promote it. Also, we'll soon be making payment links more visible on sandbox accounts so i hope that'll also help with adoption. Yet there's still a technical aspect to it - in knowing that you don't need to integrate. Possibly it could go in the Quick start guide - since that launches straight into API integration, which is a little misleading (especially if you click directly into that, rather than reading through the content chronologically - which i guess is what a lot of people do).

bahmady:
@katiebates I've made the changes per your feedback to the landing page. 

Re. payment links: that's fine, and makes sense. Makes it easier for me, too, but we can use other PRs to move that work forward/make the docs better.

PR #176, opened 2018-10-15
Improve "Before you start" section and add information about payment links

bahmady:
### Context
The "Before you start" section breaks GOV.UK style slightly, in places. 

Separately, I'm also not entirely convinced we actually need this section. I've proposed moving some of the content to the "Testing" section in this PR, but we could move the rest of the content to the landing page and more strongly emphasise the features/product page on that. I'm not sure about having an explicit "Integration" section. 

I would also suggest that we either rename the "Quick start guide" section to "API quick start guide" (or at least better reflect that that section is specifically about using the API) or that we rework that section to make it more generalised than just the API. 

1. Moves some of the content from "Before you start" to the "Testing GOV.UK Pay" section. The content in question is about testing different aspects of the platform so I think it would be better suited there.
2. Changes some language to conform to GOV.UK style and reworks the language. 
3. Deletes the "How long will it take?" section. I think having one is fine but in the form it's in at the moment (i.e. prior to this PR), it's not helpful to give an example that will only be directly applicable to a few people. 
4. Adds information about payment links. This is partly duplicated from that section: I'm actually not convinced that's a bad thing but I'll think about it.

### DIFF
--- a/source/before_you_start/index.html.md.erb
+++ b/source/before_you_start/index.html.md.erb
@@ -7,34 +7,21 @@ weight: 30
 
 Before you start testing GOV.UK Pay, you should:
 
-* read the ["Get started" guide](https://www.payments.service.gov.uk/getstarted/) and decided that GOV.UK Pay is right for your service
-* sign up for [the GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/create-service/register)
-* read the government guidance on [deploying new software](https://www.gov.uk/service-manual/making-software/deployment.html).
+* read the ["Get started"
+guide](https://www.payments.service.gov.uk/getstarted/) to decide if
+GOV.UK Pay is right for your service 
+* sign up for [the GOV.UK Pay admin
+tool](https://selfservice.payments.service.gov.uk/create-service/register) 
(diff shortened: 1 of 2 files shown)


bahmady:
Need to make a note of this:

The reusable link that’s generated. For example: https://products.pymnt.uk/pay/07e4f383aeea41d295d07de37a4dae28

Will continue to work. But when you click that link it will redirect to a payment that will expire in 90 min

bahmady:
@katiebates and I have talked about this and we're in agreement that we shouldn't have both sections.

This PR is really just style work rather than new content so we could approve this as a temporary solution since the work is already done, and I can start work on reworking the 'Before you start' and 'Quick start guide' sections into something that makes more sense.

katiebates:
> @katiebates and I have talked about this and we're in agreement that we shouldn't have both sections.
> 
> This PR is really just style work rather than new content so we could approve this as a temporary solution since the work is already done, and I can start work on reworking the 'Before you start' and 'Quick start guide' sections into something that makes more sense @katiebates? Otherwise I'm happy to just scrap this.

Sure - happy to go with that approach.

PR #179, opened 2018-10-16
Corporate surcharge

bahmady:
### Context
Adds some documentation for corporate surcharging in "Optional features" section to be added when that work is completed 

* Adds to list of optional features and reorders that list slightly
* Adds instructions as given by @alexbishop1 for adding corporate surcharges 

### Guidance to review
I've put up a test environment: https://govukpay-corporatesurcharge.cloudapps.digital/optional_features/#optional-features


source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,41 @@
...
+You can sign in to the [GOV.UK Pay admin
+tool](https://selfservice.payments.service.gov.uk/login) to see the
+payment amount and the surcharge amount listed separately. 
+
+If a corporate surcharge is applied to a particular payment, you can see this
+in API responses. The `"corporate_card_surcharge"` and `"total _amount"`
+fields contain amounts in pence. 
+
+If you issue a refund, you can optionally refund up to the full amount -
+including the surcharge. You can also use the partial refund feature to only
```
alexbishop1: I’m not sure “optionally refund up to the full amount” gets across that this is the default behaviour and I think we want to make sure we get across that the `amount_available` and `refund_amount_available` fields in the API also include the surcharge.

Maybe something like this? “If a payment had a surcharge applied, the full amount available for refund includes the surcharge. You can also use the partial refund feature to only …”

Maybe we should also update the refunding payments page to indicate that the `amount_available` and `refund_amount_available` fields include any surcharge.

source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,41 @@
...
+refund the non-surcharge total. You can read more in the ("Refunding
+payments")[/refunding_payments] section.
```
alexbishop1: This link doesn’t seem to have worked

bahmady:
Thanks @alexbishop1, I think I dropped the ball with this one so thanks for picking these up

bahmady:
Thanks for picking these up @alexbishop1, apologies for dropping the ball a bit :) 

Changes in latest commits summarised per suggestions/other things I've noticed:

* more active phrasing a la “You can set a different corporate surcharge for each of the following 4 types of card:”
* fixed prepaid/pre-paid inconsistency
* changed language re. how GOV.UK checks against numbers rather than use term "list"
* fixed extraneous space/broken link/full stop error
* changed language re. different types of amount to bulleted list, and added info about `"amount"` amount 
* changed language re. "you can use ... partial refund..."

source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,46 @@
...
+You can set a different corporate surcharge for each of the following 4
+corporate card types: 
+
+* non-prepaid credit cards 
+* non-prepaid debit cards
+* prepaid credit cards
+* prepaid debit cards 
+
+When a user enters a card number in their payment journey, the GOV.UK Pay
+platform checks it against the 4 corporate card types. If a surcharge is
```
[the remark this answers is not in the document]
bahmady: Yeah, you're right: "checks it against" doesn't sound right, and still sounds like there's some sort of implied list. 

Your suggestion sounds good.

source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,46 @@
...
+
+You can sign in to the [GOV.UK Pay admin
+tool](https://selfservice.payments.service.gov.uk/login) [external link] to see the
+payment amount and the surcharge amount listed separately. 
+
+If a corporate surcharge is applied to a particular payment, you can see this
+in API responses:
+
+* the `"total_amount"` field contains the total amount including any surcharge
+* the `"amount"` field contains the amount without any surcharge
```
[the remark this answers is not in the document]
bahmady: Does `"corporate_surcharge"` only appear if there _is_ a corporate surcharge?

source/refunding_payments/index.html.md.erb
```diff
@@ -62,7 +67,10 @@ If you needed to know the details of the partial refunds (for example, whether t
 
 ## Specifying the expected refund available
 
-When you submit a refund request for a payment via the API, you can optionally specify a ``refund_amount_available`` value in the body of the request.
+When you submit a refund request for a payment via the API, you can optionally
+specify a ``refund_amount_available`` value in the body of the request. This
+value [includes any corporate card 
+surcharges](/optional_features/corporate_card_surcharges/#corporate-card-surcharges).
```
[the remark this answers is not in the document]
bahmady: It's a bit tricky without rewording this whole section (which I will do eventually), because it breaks the flow of the other paragraphs a little. I wonder if it might be best to put the note about adding corporate surcharges to the bottom of the "Specifying the expected..." section. 

Just to be 100% clear: we are talking about the `"refund_amount_available"` value in the context of including corporate card surcharges [in something]?

bahmady:
@alexbishop1 I've followed the discussions in the individual comments to try to improve this, per your suggestions. The only extra thing worth mentioning is that I've had to be a bit creative to fit in with GOV.UK style in one place—rather than bullets for the responses, I'm trying h3s. 

Let me know what you think (or please feel free to tell me I need to talk to someone else; I don't want to take too much of your time). There's no rush for this in any case.

Thanks again

source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,57 @@
...
+You can set a different corporate surcharge for each of the following 4
+corporate card types: 
+
+* non-prepaid credit cards 
+* non-prepaid debit cards
+* prepaid credit cards
+* prepaid debit cards 
+
+When a user enters a card number in their payment journey, the GOV.UK Pay
+platform checks to determine if it's one of the 4 corporate card types. If a
```
alexbishop1: Curly apostrophe ’ here please

source/optional_features/corporate_card_surcharges/index.html.md.erb
```diff
@@ -0,0 +1,57 @@
...
+You can sign in to the [GOV.UK Pay admin
+tool](https://selfservice.payments.service.gov.uk/login) (link opens in new
+window) to see the payment amount and the surcharge amount listed separately. 
+
+If a payment has a surcharge applied, the full amount available for refund
+includes the surcharge. You can also use the [partial refund
+feature](/refunding_payments/#payment-refund-status-and-partial-refunds) to
+only refund the total amount excluding the surcharge. 
+
+If a corporate surcharge is applied to a particular payment, you can see this
```
alexbishop1: Similar to the comment above, we probably want this to be “corporate card surcharge”

bahmady:
I've changed the h3s to h4s ... the h3s looked a bit weird.

katiebates:
All looks good to me. I've raised the question on slack about whether we should hint that we could apply percentage if needed. But happy to go with this for now.

katiebates:
On reflection, it would be good if we can find a way to soften this "Each surcharge is a flat amount added to a payment, not a percentage." We visited an LA yesterday who want to use corporate surcharging but with percentages - and i think this is probably quite a common use case, as card schemes do charge on a % basis so it makes sense that a merchant just passes on this charge.

bahmady:
@katiebates How does it look now? I've updated in the PR and this test deployment:

https://govukpay-corporatesurcharge.cloudapps.digital/optional_features/#optional-features

bahmady:
I'm also not sure when to merge this as I know this feature technically isn't live yet so do please let me know when it's OK to do that.

bahmady:
Based on good feedback I've changed h4s at the bottom of the doc to a table so this may also be worth reviewing. The other edits I'll make shortly are just minor language ones.

katiebates:
> I'm also not sure when to merge this as I know this feature technically isn't live yet so do please let me know when it's OK to do that.

You can merge it whenever you're ready - the majority of the work is finished and in any case, it can't be 'live' for anyone until we manually enable it anyway.

katiebates:
one minor thing - the docs say that the user will be informed about the surcharge. do we need to be a little clearer how this happens? (e.g. on the payment page, when the user enters the card number..?)

bahmady:
@katiebates I've added this line:

> If a surcharge is applicable, when users enter their card details on the payment
page, the platform will inform them that a surcharge of a certain amount
will be added to the total.

It's a bit meaty but I can't think of a better solution.

bahmady:
@katiebates In hindsight I'm not sure we do need this edit, since we have this in the preceding sentence:

>When a user enters a card number in their payment journey...

I'm not sure, though.

PR #185, opened 2018-10-25
Account structure

bahmady:
### Context
Draft version of the account structure documentation—subject to change. 

Adds draft version of account structure docs. Needs Conor to work on a new diagram(?)

### Guidance to review
Test deployment: https://govukpay-accountstructure.cloudapps.digital/account_structure/#account-structure


katiebates:
will add more comments later, but just wanted to make a note (from zendesk analysis) that we _may_ want to add something here (or in another section) about account structure and destination funds. eg if you want your money to go into separate bank accounts then you'll need different accounts (and different MID codes) etc. ticket #3166291

bahmady:
@katiebates I've attempted to tidy this up a bit. Now posted to https://govukpay-accountstructure.cloudapps.digital/account_structure/#account-structure

Let me know if it's going in the right direction :)

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+* email template to user
+* accepted card types
+* name that appears on the payment pages 
+* accept 3DS 
+
+## Editing PSP services 
+
+Within each service for the PSP, you can edit: 
+
+* what appears on bank statement of your users
```
katiebates: i'm not totally sure this is true - @tillwirth can you confirm?

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+![Account structures diagram](https://raw.githubusercontent.com/alphagov/pay-tech-docs/account_structure/source/images/account_structure.png)
+
+You may have 2 services, for example for parking permits and parking fines.
+The diagram takes this example and illustrates the different possible
+structures you can have:
+
+### Type 1 (GOV.UK Pay separate, PSP separate)
+
+"Parking permits" and "Parking fines" are considered as separate services in
+both GOV.UK Pay and the PSP.
```
katiebates: Might want to add that they can be paid into separate bank accounts since they also have their own PSP accounts.

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+
+### Type 2 (GOV.UK Pay combined, PSP combined)
+
+"Parking permits" and "Parking fines" are considered as a single combined
+service in GOV.UK Pay and in the PSP.
+
+### Type 3 (GOV.UK Pay separate, PSP combined)
+
+"Parking permits" and "Parking fines" are considered as separate services in
+GOV.UK Pay and as a single combined service in the PSP.
```
katiebates: Destination bank account will be the same for both services.

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,51 @@
...
+service in GOV.UK Pay and in the PSP.
+
+### Type 3 (GOV.UK Pay separate, PSP combined)
+
+"Parking permits" and "Parking fines" are considered as separate services in
+GOV.UK Pay and as a single combined service in the PSP. 
+
+## Editing GOV.UK Pay services 
+
+Each service defined in GOV.UK Pay has its own API.
```
katiebates: is it right to say 'own API' - or own API key?

katiebates:
Definitely going in the right direction! I think we could launch it and collect user feedback on it.

source/account_structure/index.html.md.erb
bahmady: I was confused about this—this is from the copy in the original slides from Miriam I think? I don't think this is an accurate statement.

bahmady:
OK, I need writer review on this: I'm going to create a GitHub label that indicates that I still need that. It won't be as much of a blocker as before, though, since I have a new system.

bahmady:
@katiebates do you want to publish this with the existing diagram, then update when we have a new one from Conor?

source/account_structure/index.html.md.erb
bahmady: Done

source/account_structure/index.html.md.erb
bahmady: Done 👍 and for Type 2

bahmady:
One thing that strikes me is that we don't actually explain how to do any of the things we're saying we can edit...

source/account_structure/index.html.md.erb
alexbishop1: The only thing I can think this is intended to be is “API keys” or something to convey that each service is separated at the API level i.e. you only ever work with one service at a time via the API. (In actuality, the separation is at the gateway account level: in our system, there are services and each service has one or more gateway accounts — usually a live account and a test sandbox account — and one works with a gateway account at the API level. However, in user-facing docs, we tend to only use the term “gateway account” when talking specifically about a gateway account with a PSP, eliding our sense of gateway accounts into services.)

source/account_structure/index.html.md.erb
katiebates: oh i meant about customising what's on the bank statement! sorry i wasn't clear.

bahmady: Just a note for completeness: this should be OK as it stands

source/account_structure/index.html.md.erb
bahmady: I think in that case (thank you @alexbishop1), bearing it mind that the gateway account interaction is implied, is it sufficient to simply say(?):

> Each service defined in GOV.UK Pay has its own API key.

Or do we need to be more specific about how each service is separated at the gateway account level, in addition?

bahmady:
I've updated this based on feedback and comments, and have updated https://govukpay-accountstructure.cloudapps.digital/account_structure/#account-structure

cc @katiebates

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,80 @@
...
+As an example, take 2 services called "parking permits" and "parking fines": 
+
+### Type 1 (GOV.UK Pay separate, PSP separate)
+
+![Account structures diagram](https://raw.githubusercontent.com/alphagov/pay-tech-docs/account_structure/source/images/accountstructure_1.png)
+
+"Parking permits" and "Parking fines" are considered as separate services in
+both GOV.UK Pay and the PSP. 
+
+This is the default configuration for services.
```
katiebates: not sure about this line

source/account_structure/index.html.md.erb
```diff
@@ -0,0 +1,80 @@
...
+
+### Type 2 (GOV.UK Pay combined, PSP combined)
+
+![Account structures diagram](https://raw.githubusercontent.com/alphagov/pay-tech-docs/account_structure/source/images/accountstructure_2.png)
+
+"Parking permits" and "Parking fines" are considered as a single combined
+service in GOV.UK Pay and in the PSP. 
+
+You may want to use this configuration if you want to simplify the
+administration of multiple services in GOV.UK Pay, and have them appear as a
```
katiebates: "simplify the administration of multiple services in GOV.UK Pay" - yep, this describes it well. basically, they'll get all their transactions in one place - in the same transactions list and in the same CSV. they might need to configure their reference numbers differently eg parkingfine1234 and parkingpermit1234 if they need to separate out payments. 

"appear as a single item.." - i don't think this is really true. it'll appear as a single item regardless of the service setup

PR #189, opened 2018-10-31
Improves 'Refunds' section, adds note about searching for refunds with the API, adds note about refunds to 'Reporting' section

bahmady:
### Context
Getting some information about searching for refunds with the /refunds endpoint in the public documentation

Improves some language in refunding before a proper revamp, adds notes about refunding to reporting page, adds notes about searching for refunds with the API

### Guidance to review
Live preview: https://govukpay-refund-revamp.cloudapps.digital/refunding_payments/#refunding-payments

### DIFF
--- a/source/refunding_payments/index.html.md.erb
+++ b/source/refunding_payments/index.html.md.erb
@@ -265,7 +265,7 @@ refund exists, the `"status"` field will contain one of:
 `”success”`
 `”error”`
 
-## Generate a list of refunds (search refunds)
+### Generate a list of refunds (search refunds)
 
 `GET /v1/refunds`
 
--- a/source/refunding_payments/index.html.md.erb
+++ b/source/refunding_payments/index.html.md.erb
@@ -5,9 +5,11 @@ weight: 80
 
 # Refunding payments
 
-GOV.UK Pay now supports refunding payments. You can choose to refund part of a payment, or the full amount.
+GOV.UK Pay supports refunding payments. You can choose to refund part of a
+payment, or the full amount.
 
-After issuing a partial refund of a payment, you can issue further partial refunds, until the full amount of the original payment has been refunded.
+After issuing a partial refund of a payment, you can issue further partial
+refunds, until the full amount of the original payment has been refunded.
 
 Each payment has a refund status which can take one of the following values:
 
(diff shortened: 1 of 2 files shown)


katiebates:
can we also change the refunds notification section please? We now send an email which confirms that a refund has been successfully processed (check with Silvia - i think we don't send one if the refund fails). The refund email can be switched on/off in the email settings part of the admin site.

source/refunding_payments/index.html.md.erb
```diff
@@ -181,34 +197,110 @@ Initially, in a live environment, the status returned will be ``submitted``. Aft
...
+or via the API. You should arrange to notify end users about payment failures
+and refunds as appropriate for your service.
 
-In the details view, you can use the red **Refund payment** button at the upper right to carry out a full or partial payment.
+## Refund destination
 
-![](images/refund-payment-button.png)
+Refunds are sent back to the source account of the original payment. It is not
+possible to refund a payment to another card or bank account. If the original
+card expired or was canceled, the bank will attempt to credit the customer’s
```
katiebates: cancelled rather than canceled?

katiebates:
the screenshot of the transactions page is out of date, since the changes made to the search filters. 

also, in the refunding from admin site section: "In this list, you can select an individual payment to see details of that payment, including any previous refunds." This is true, but it might also be worth mentioning that any refunds also appear in the transactions list as single line items too. So if i paid £100, and then got a refund for £50 and £30, then there would be three entries in the transaction list. It works the same in the CSV too. Or it might be sufficiently visible/ self-explanatory.

katiebates:
the refunds endpoint section looks good to me.

katiebates:
i know this is pre-work before a full revamp so using this as an opportunity to throw in my twopence: I wonder if the "Specifying the expected refund available" section should come after, or as a subset of the "Refunding with the API section" since it's _part_ of refunding via the API. Also "Handling refund errors" looks like it could be a sub-section of "Refunding with the API" too?

katiebates:
given that the PR does what it says on the tin, i'm approving and merging it. the other comments can be addressed in separate PRs.

bahmady:
I used this PR as an opportunity to tidy up the language in the main section, since I noticed a few errors and style problems, but yes, I missed a few. I'll open a new PR today referencing this one to address the wider rework of the refunds section.

PR #204, opened 2018-11-08
Diagrams

bahmady:
### Context
The 'Architecture' section in the existing live tech docs contains orphaned diagrams with no supporting text. We should find homes for them or create them. 

* Temporarily removes the integration architecture diagram (to be reintroduced when I've created new guidance with David) from 'Architecture'.

* Replaces the state machine diagram which @katiebates has expressed may be inaccurate/misleading in the 'Payment flow' section with the payment architecture diagram from 'Architecture'. 

NOTE: If we do accept these proposed changes then I'll probably have to make further changes at a later date when new diagrams guidance is released.

*Removes the 'Architecture' section altogether.

### Guidance to review
https://govukpay-diagramschanges.cloudapps.digital


bahmady:
@katiebates how does this look now? https://govukpay-diagramschanges.cloudapps.digital/payment_flow_overview/

bahmady:
Actually @katiebates I'm going to make this a bigger PR with improvements elsewhere so let's not merge just yet

katiebates:
"The following diagram illustrates how a payment can result in the different possible final states:"

It also illustrates the other states that the payment passes through before reaching a final state... these states are relevant to services as they're exposed too. also i wonder about the use of "can" since the payment will always reach a terminal state

bahmady:
Have amended that text

PR #209, opened 2018-11-19
Add a note explaining that payment links are only available on live accounts

rhowe-gds:
### Context
We sometimes get queries about why the payment links feature isn't visible. This is because the options don't show up for non-live accounts.

Add payment links to the list of differences between live and non-live accounts

### Guidance to review
Make sure my statement is actually true


katiebates:
i'd rather we leave this out - it's possible to access them using the direct URL on test accounts and we often tell potential partners to do this. and we'll change it soon so they are visible on test accounts.

PR #210, opened 2018-11-23
'Use your own payment failure pages'

bahmady:
### Context
Introduction of 'Custom error pages' content (directly redirect) 

Adds 'Custom error pages' content in 'Optional features' section

### Guidance to review
In the first instance I'd just like to know whether this is on the right lines and whether we need to do anything else. I can rebase or `--reset soft` and amend the commits after to ensure it's only one commit that goes into the main repo when this is done. Also, I'll need to have this writer-reviewed.

### DIFF
--- a/source/optional_features/use_your_own_error_pages/index.html.md.erb
+++ b/source/optional_features/use_your_own_error_pages/index.html.md.erb
@@ -1,0 +1,52 @@
+---
+title: Use your own payment failure pages
+weight: 135
+---
+
+# Use your own payment failure pages
+
+You can [contact us](/support_contact_and_more_information/#contact-us) to set
+this feature up for your service.
+
+If you do not use your own payment failure pages, users will see standard
+GOV.UK Pay payment failure pages.
+
+Terminal, unsuccessful payment states are:
+
+* payment declined
+* payment cancelled
+* payment expired
+* error (for example, technical error)
+
+If you use this feature, GOV.UK Pay will immediately redirect the user to the
+`return_url` when a payment reaches a terminal unsuccessful
+state. You can [set the `return_url` manually for your service](/payment_flow//#making-a-payment) in the request
+to the `POST /v1/payments` API call.
+
+You should configure your service to query payments to determine what
+payment failure states are from API responses. Once your service
+knows the [state of a payment](/payment_flow/#check-the-status-of-a-payment), it should direct users appropriately. For
+example if a user's payment expires, you may want your service to return the
+user to the start of their payment journey.
+
+You can experiment with querying payments using [our
+interactive API
+explorer](https://gds-payments.gelato.io/api-explorer/gov-uk-pay-api/versions/1.0.2//)
+[external link].
+
+In very rare cases, GOV.UK Pay is unable to redirect payments. You should
+[contact us](/support_contact_and_more_information/#contact-us) if this
+happens.
+
+## Expired payments
+
+If a user abandons a payment and it expires, GOV.UK Pay can only redirect the
+user if they did not leave the payment screen. If the user tries to
+continue their journey from the payment screen after their payment has
+expired, they will be returned to the service.
+
+You may want to read about [what happens when a user does not complete
+their payment
+journey](/making_payments/#when-a-user-does-not-complete-their-payment-journey).
+
+
--- a/source/optional_features/index.html.md.erb
+++ b/source/optional_features/index.html.md.erb
@@ -11,3 +11,4 @@ GOV.UK Pay supports some optional features which need configuration:
 * [Welsh language](/optional_features/welsh_language/#welsh-language-payment-pages) of payment pages
 * [Delayed capture](/optional_features/delayed_capture/#delayed-capture) of payments
 * [Corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) added to payments
+* [Using your own payment failure pages](/optional_features/use_your_own_error_pages) for your service


source/optional_features/custom_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,24 @@
+---
+title: Custom error pages 
+weight: 135
+---
+
+# Custom error pages
+
+GOV.UK Pay supports custom error pages.
```
katiebates: I'd prefer this to be more explicit about the behaviour that the user will see, something along the lines of, "you can use your own pages for unsuccessful payments"

source/optional_features/custom_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,24 @@
+---
+title: Custom error pages 
+weight: 135
+---
+
+# Custom error pages
```
katiebates: I don't think this is the right title. 'Custom' suggests you're customising something provided by the GOV.UK Pay service - which isn't the case. Also, in future we may offer the option of genuinely customising our unsuccessful payment pages (for example, allowing a service to have an optional paragraph) so this wouldn't leave us many options should we do that. We should have a look at the language that services which have requested this feature have used when talking about it.

source/optional_features/custom_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,24 @@
...
+GOV.UK Pay supports custom error pages.
+
+When a payment reaches a terminal, unsuccessful state, GOV.UK Pay will
+immediately redirect the user to the return URL (`return_url`), which you can
+set in the response to the `POST /v1/payments` API call. 
+
+To set this up for your service, you should make a query to the GOV.UK Pay API
+to determine what the failure state is, and set an appropriate error message. 
+
+If you do not use custom error pages, users will see generic unsuccessful
```
katiebates: not sure about using 'generic' here - it could suggest that Pay only has one generic unsuccessful payment screen.

source/optional_features/custom_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,24 @@
...
+
+# Custom error pages
+
+GOV.UK Pay supports custom error pages.
+
+When a payment reaches a terminal, unsuccessful state, GOV.UK Pay will
+immediately redirect the user to the return URL (`return_url`), which you can
+set in the response to the `POST /v1/payments` API call. 
+
+To set this up for your service, you should make a query to the GOV.UK Pay API
```
katiebates: this doesn't quite describe how a service would set it up - they need to contact us for the feature to be enabled. also, regardless of whether this feature is used, a service would still make a GET request per transaction and (ideally) should parse the failure state so they can direct the user on the right journey. For example, if the user cancelled the payment, the service might determine that the user shouldn't be taken back to the start of a payment journey again, whereas if the payment expired, they might.

source/optional_features/custom_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,24 @@
...
+to determine what the failure state is, and set an appropriate error message. 
+
+If you do not use custom error pages, users will see generic unsuccessful
+payment screens.
+
+## Expired payments 
+
+If the user abandons a payment and it expires, GOV.UK Pay will only redirect
+the user if they did not actually leave the payment screen. If they try to
+continue it, they will be returned to the service.
```
katiebates: might want to mention here that this is why we still expect the service to have a clean-up job if they use this feature

source/optional_features/custom_error_pages/index.html.md.erb
bahmady: Well both you and Alex used the term "Use your own error pages" so perhaps we should try that?

source/optional_features/custom_error_pages/index.html.md.erb
bahmady: That's fine—as I suggest in the other comment there's no real reason why we shouldn't use that as the title for the section.

source/optional_features/custom_error_pages/index.html.md.erb
bahmady: Can we reasonably expect the service (reader of the documentation) to know what will be meant by the term "clean-up job", and how to use it? Are any other terms likely to be used?

source/optional_features/custom_error_pages/index.html.md.erb
bahmady: From Alex: want to get across that the pages are "generic" in the sense that, regardless of what your service is you'll have similar error pages — if the service does their own error pages, they're in a good place to tell the user what to do with a specific problem. 

Could use the term "standard GOV.UK Pay error pages" — these are generally good, but services can make them more specific to the problem(s) at hand, depending on circumstances.

source/optional_features/custom_error_pages/index.html.md.erb
katiebates: something like that, although i'd be careful with the word 'error' because that specifically describes one of the scenarios rather than the entire set. the scenarios are: 
1. payment declined 
2. payment cancelled 
3. payment expired 
4. error (eg technical error)
One service who requested it have called it 'payment failure screens', which (IMO) has a better ring to it than 'unsuccessful payment'.

source/optional_features/custom_error_pages/index.html.md.erb
katiebates: 'standard' works; see my comment about 'error' above

source/optional_features/custom_error_pages/index.html.md.erb
katiebates: mop up job ;)

bahmady:
@katiebates I've tried to incorporate all yours and @alexbishop1's feedback and I hope this is better.

I'm not sure about the section we have on 'expired payments' because we don't single out any of the other possible terminal unsuccessful states.

source/optional_features/use_your_own_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,45 @@
...
+* payment declined
+* payment cancelled
+* payment expired
+* error (eg technical error)
+
+If you use this feature, when a payment reaches one of these states, GOV.UK
+Pay will immediately redirect the user to the return URL (`return_url`). You
+can set the `return_url` manually for your service, in the response to the
+`POST /v1/payments` API call. 
+
```
katiebates: oh, i thought it was set in the request rather than response? @alexbishop1 can confirm

bahmady: Yeah it will be, this is just a mistake. I'll wait for Alex' reply in any case.

katiebates:
> @katiebates I've tried to incorporate all yours and @alexbishop1's feedback and I hope this is better.
> 
> I'm not sure about the section we have on 'expired payments' because we don't single out any of the other possible terminal unsuccessful states.

we might also want to single out Error - basically, in most categories of error (@alexbishop1 to confirm), we probably won't be able to redirect the user back to the service

bahmady:
> we might also want to single out Error - basically, in most categories of error (@alexbishop1 to confirm), we probably won't be able to redirect the user back to the service

OK, makes sense—but we can for the other categories (declined, cancelled)?

source/optional_features/use_your_own_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,45 @@
...
+You should configure your service to query payment states and determine what
+the failure state is. Your service should then direct users appropriately. For
```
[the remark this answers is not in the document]
bahmady: We could just say "query the payment [ID]" in the API and link to the API reference / explorer? Unless we need to be explicit.

bahmady:
@katiebates I've had another try: I'd like to go through the entire "Payment flow overview" section and improve that as a separate task. So far I've only done a cosmetic review of it.

source/optional_features/use_your_own_error_pages/index.html.md.erb
```diff
@@ -0,0 +1,50 @@
...
+this feature up for your service. 
+
+Terminal, unsuccessful payment states are:
+
+* payment declined
+* payment cancelled
+* payment expired
+* error (eg technical error)
+
+If you use this feature, when a payment reaches one of these states, GOV.UK
```
katiebates: in some categories of error, we won't be able to redirect. however this is extremely rare (most likely is NAXSI errors). in routine errors and payment failures we'll be able to redirect. so it's probably fine to leave it like this!

bahmady: There's probably no real problem adding that in rare cases we won't be able to redirect, as long as you're OK adding a note to "Contact us" if that happens?

katiebates: sure

PR #213, opened 2018-11-28
Billing address note 

bahmady:
### Context
We should include that if you opt not to collect billing address, then the billing address fields will not be returned in public API when doing a GET request.

Adds warning text about billing address: if not set, address fields will not be returned in public API when doing a GET request

### Guidance to review
https://govukpaytest.cloudapps.digital/payment_flow/#overview

We also had this note in the Trello card but I'm not sure about including this:

> we might also want to point out that if you switch on/off collecting billing address, it can take up to 45 minutes for the change to be seen on the payment pages due to caching

### DIFF
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -35,6 +35,14 @@ When an end user needs to make a payment, your service makes an API call to crea
 
 The end user enters their payment details (for example, credit/debit card details and billing address) on the GOV.UK Pay pages. GOV.UK Pay verifies the payment with the underlying PSP. 
 
+<div class="govuk-warning-text">
+  <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+  <strong class="govuk-warning-text__text">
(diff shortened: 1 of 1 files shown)


katiebates:
Sorry Ben, after i wrote the Trello card I discovered a bit more about this but forgot to update the trello :( here's the jira story which describes the behaviour https://payments-platform.atlassian.net/browse/PP-4403 
Based on that, it's not so much that the address fields aren't returned, it's that after switching off billing address, we'll return a null response. 

I agree it's not the right place to put a note about it not taking effect until up to 45 minutes later but it would be good to have it somewhere, in case a service team are thinking about using the feature and then confused about why it's not taken effect (plus they might want to make the change out of hours). Usually changes made in the admin site take effect immediately so it's atypical. I think it's worth having a small section on Billing address (probably in Optional features) since - although it's in the admin site - it does have technical impacts.

bahmady:
No problem @katiebates: I'll close this PR and add a card on Trello for creating some billing address docs?

katiebates:
sounds good!

PR #214, opened 2018-11-28
Remove trailing whitespace in all index.html.md.erb files

bahmady:
### Context
There are a lot of trailing whitespaces in `index.html.md.erb` files we use to render our docs. This isn't good practice even though it's not actually important as far as Middleman's rendering is concerned, because it may be awkward for some people navigating through the docs in future when making edits using Git. 

Rids all docs files of whitespace. 

### Guidance to review
I've tested and built this locally and can't see any changes as far as the user is concerned. If I've missed something we can revert this commit.

### DIFF
--- a/source/account_structure/index.html.md.erb
+++ b/source/account_structure/index.html.md.erb
@@ -1,41 +1,41 @@
 ---
-title: Account structure 
+title: Account structure
 weight: 95
 ---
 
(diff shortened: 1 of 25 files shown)


bahmady:
I'm glad I remembered how to `rebase` 🤦‍♂️

PR #218, opened 2018-11-29
Worldpay docs inaccuracies 

bahmady:
### Context
We haven't had access to the user journey for switching to live on Worldpay until now, so there are some inaccuracies in the documentation. It seems like Worldpay may have amended/updated some terminology, as well. Also, we should be recommending __New Username__ above __Original XML Username__. 

Amends Worldpay STL documentation to reflect actual user journey/Worldpay updates

### DIFF
--- a/source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
+++ b/source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
@@ -34,27 +34,22 @@ If __Switch to Production__ is at the bottom of the left-hand menu, toggle it on
 To set up your profile, you need to copy credentials from your Worldpay account
 into your GOV.UK Pay account.
 
-You will need both your [Worldpay
+You must separately sign in to both your [Worldpay
 account](https://secure.worldpay.com/sso/public/auth/login.html) and your
-[GOV.UK Pay account](https://selfservice.payments.service.gov.uk/login) open
-and signed in, in different web browser tabs.
+[GOV.UK Pay account](https://selfservice.payments.service.gov.uk/login). You
+should have both accounts open, in different web browser tabs.
 
-1. Select __Profile__ in the left-hand menu in your Worldpay account.
+1. In your Worldpay browser tab, select __Profile__ in the left-hand menu. If you
+are prompted to choose a merchant code, select the appropriate one.
 
 1. In your GOV.UK Pay browser tab, go to __My services__ to select the live
 service you want to set up. Select __Settings__, then __Account credentials__,
 then __Edit credentials__.
 
-1. In your Worldpay browser tab, select __Account__ in the left-hand
-menu.
-
-1. From the __Identification__ section, copy the Worldpay __Merchant Code__ into
+1. In Worldpay, copy the Worldpay __Merchant Code__ from __Identification__ into
 the GOV.UK Pay __Merchant ID__ field.
 
-1. Select the checkbox __Enable original username__ and select __save__.
-
-1. Copy the Worldpay __Original XML Username__ into the GOV.UK Pay
-__Username__ field.
+1. Copy the Worldpay __New Username__ into the GOV.UK Pay __Username__ field.
 
 1. Change the Worldpay XML password:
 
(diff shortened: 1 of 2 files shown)


source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
```diff
@@ -34,27 +34,25 @@ If __Switch to Production__ is at the bottom of the left-hand menu, toggle it on
...
 
-1. In your Worldpay browser tab, select __Account__ in the left-hand
-menu.
-
-1. From the __Identification__ section, copy the Worldpay __Merchant Code__ into
+1. In Worldpay, copy the Worldpay __Merchant Code__ from __Identification__ into
 the GOV.UK Pay __Merchant ID__ field.
 
-1. Select the checkbox __Enable original username__ and select __save__.
+1. In Worldpay, make sure the checkbox __Enable original username__ is toggled. Select __save__.
```
katiebates: what exactly does toggled mean in this case - that it shouldn't be be selected? (it shouldn't be, as the new username should be used)

bahmady: I miswrote this. I think the following would be better—let me know your thoughts:

> In Worldpay, tick __Enable original username__ if it is not already ticked, and then __Save Profile__.

katiebates: but we want them to use the New Username instead, so they shouldn't tick that box

bahmady: Ah of course, I'm sorry, too many things on my mind at the same time. We should delete this sentence altogether, I think. Are you happy with this line(?):

> If you are sure you need the __Original XML Username__, use that instead.

bahmady:
@katiebates added some stuff from Miriam and Till I've received via email today and snuck in a rider about moving the placement of the 'What to do before you switch to live...' content line

source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
```diff
@@ -34,27 +34,22 @@ If __Switch to Production__ is at the bottom of the left-hand menu, toggle it on
...
 1. In your GOV.UK Pay browser tab, go to __My services__ to select the live
 service you want to set up. Select __Settings__, then __Account credentials__,
 then __Edit credentials__.
 
-1. In your Worldpay browser tab, select __Account__ in the left-hand
-menu.
-
-1. From the __Identification__ section, copy the Worldpay __Merchant Code__ into
+1. In Worldpay, copy the Worldpay __Merchant Code__ from __Identification__ into
 the GOV.UK Pay __Merchant ID__ field.
```
katiebates: we're just about to change this to Merchant Code in GOV.UK Pay to make it less confusing, sorry, should've mentioned before!

katiebates: i'll ping you when i sign this off, then it can be changed then so it doesn't hold this up

source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
katiebates: no worries! i agree it's easier this way. 
I think if we include the line about Original XML username, we should make it clear that that you'll need to tick the box that says "Enable Original XML username" and click save. i can't really think of a reason they'd need to use it though.

source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
katiebates: another thing for a future mini PR, is to say when you test it using a payment link, to open that up in another window because it's not particularly easy to navigate back to your admin site account once you've gone on the payment link journey

source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
bahmady: I agree: I think it's better without

bahmady: As it happens I deleted even the "If you are sure..." line as I guess you noticed but I think we're probably OK?

source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
bahmady: Sure, not sure where to put that info but happy if you want to put it in a Trello card or just email me/add a comment here and I can open a new PR?

bahmady: > we're just about to change this to Merchant Code in GOV.UK Pay to make it less confusing, sorry, should've mentioned before!

No problem, just ping me as you said

PR #221, opened 2018-12-03
Reworks the 'Testing' section

bahmady:
### Context
The 'Testing' section breaks style in some places

Language rework per above

### DIFF
--- a/source/testing_govuk_pay/index.html.md.erb
+++ b/source/testing_govuk_pay/index.html.md.erb
@@ -5,18 +5,27 @@ weight: 90
 
 # Testing GOV.UK Pay
 
-You will receive a test account for testing in addition to your live credentials.
+When your GOV.UK Pay account is [switched to
+live](/switching_to_live/#switching-to-live), you will still have a ‘test’
+account for general testing and experimenting with different settings and
+features.
 
-When testing, you’ll need to ensure:
+When you test, you should make sure that you’ve tested with [mock card
+numbers](/#mock-card-numbers-for-testing-purposes) to simulate both successful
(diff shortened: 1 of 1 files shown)


source/testing_govuk_pay/index.html.md.erb
```diff
@@ -5,18 +5,25 @@ weight: 90
 
 # Testing GOV.UK Pay
 
-You will receive a test account for testing in addition to your live credentials.
+When you sign up for GOV.UK Pay, you will receive a test account in addition
+to your live credentials.
```
katiebates: I think the original wording on this was a bit confusing - i think it should be something like 'When we switch your account to live, you will still have a test account' (for testing out settings etc). Basically, you start off with a test account, we create a live account linked with the service, and you still have your original test account. Otherwise, i'm not really sure what's meant here by 'live credentials'....

source/testing_govuk_pay/index.html.md.erb
```diff
@@ -5,18 +5,25 @@ weight: 90
...
 
-When testing, you’ll need to ensure:
+When you test, you should make sure that you’ve tested with [mock card
+numbers](/#mock-card-numbers-for-testing-purposes) to simulate both successful
+and unsuccessful transactions. You must never use real card numbers for
+testing purposes, because this breaks PCI rules.
```
katiebates: i'm not totally sure this is true (about PCI), although it's probably good practice; the more important thing is that it just won't work. you have to use test cards with a test account.

bahmady:
@katiebates had a go with your suggestions, if you're happy can merge this without the need for another PR?

PR #222, opened 2018-12-03
Language/style/flow improvements to 'Making payments' section

bahmady:
### Context
The 'Making payments' section breaks style in some places, and I think it could be improved a bit in terms of language/flow

Suggestions to achieve the above 

### Guidance to review
This isn't ready to merge at time of opening; I'm going to leave some review comments so any help with these would be greatly appreciated. It does however not need a writer review.


source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,140 @@ weight: 70
...
+completed a payment using GOV.UK Pay.
 
-You will need to store the URL from the Location header or in the ``self``
-section of ``_links`` in the JSON body (the same URL is shown in both places).
-This URL contains the GOV.UK Pay ``payment_id`` which identifies the
-payment. An authenticated GET request to the URL will return information about
-the payment and its status.
+You will need to store the URL from the `Location` response header, or from
+the `self` section of `_links` in the JSON body.  This URL contains the GOV.UK
+Pay `payment_id`, which identifies the payment. You can make a GET request to
```
bahmady: The way this is worded (I haven't changed this much from the original yet) seems to suggest that `payment_id` and `reference` are the same, or at least serve the same function. Is there an obvious thing to point to to distinguish them?

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,140 @@ weight: 70
...
-intercept users’ payments.
+You can create a secure random ID,such as a UUID, and include this as part of
+the `return_url`. You should use a different ``return_url`` for each payment.
+Securely generated UUIDs are not guessable, so malicious users will
+not be able to securely generated UUID is not guessable, fraudsters will not
+be able to alter the payment ID and intercept other users’ payments.
 
->Note: If you create an ID yourself, you’ll likely need to store this in a
-datastore mapped to the payment ID just after you create a payment.
+If you create an ID yourself, you will likely need to store this in a datastore mapped
```
bahmady: Do we definitely need this section? Can we word it in such a way that it's a bit more committal?

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,140 @@ weight: 70
...
 ## Accepting returning users
 
- A user directed to the return URL could have:
+ A user directed to the `return_url` could have:
 
  - paid successfully
- - not paid because their card was rejected or they clicked cancel
- - not paid because your service (via the API) or service manager (via the self-service dashboard) cancelled the payment in progress
+ - not paid because their card was rejected or they selected __Cancel__
+ - not paid because your service (via the API) or service manager (via the GOV.UK Pay admin tool) cancelled the payment in progress
```
bahmady: Just want to check to be 100% sure that 'GOV.UK Pay admin tool' is an acceptable substitute for 'self-service dashboard'.

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,140 @@ weight: 70
...
 
-You can still check on the status of these payments by making a GET request
-using the Location Header or Self Link, the same way you would if they were
-redirected, but just after a set time (eg, an hour).
+You can still [check on the status of these payments](#making-payments), but
+after a set time (eg, an hour).
 
->Note: GOV.UK Pay will expire incomplete payments after 90 minutes, but you should expect an occasional success or failure if the user experienced problems right at the moment of the redirect.
+GOV.UK Pay will expire incomplete payments after 90 minutes, but you should
+expect occasional notifications of success or failure if users experienced
```
bahmady: I wasn't sure how to word this, based on the original.

* Do we need this information?
* Can anyone think of a better way of wording it, if we do?

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,139 @@ weight: 70
...
-fraudster could not alter the payment ID and intercept other users’ payments.
+### Create a secure random ID
 
-+ create a secure random ID (such as a UUID) and include this as part of the
-``return_url``, using a different ``return_url`` for each payment. Since a
-securely generated UUID is not guessable, fraudsters will not be able to
-intercept users’ payments.
+You can create a secure random ID,such as a UUID, and include this as part of
+the `return_url`. You should use a different ``return_url`` for each payment.
+Securely generated UUIDs are not guessable, so malicious users will
```
bahmady: Not sure about using the word 'guessable'. Could we just say something like:

> Malicious users will be unable to guess securely generated UUIDs

Is this enough? Is it implied they won't have other means of access using the UUIDs?

source/making_payments/index.html.md.erb
alexbishop1: There’s the fact that the `payment_id` is automatically generated by us whereas the `reference` is generated by the service. Can you do anything with that?

source/making_payments/index.html.md.erb
alexbishop1: The important part is that they need to have some way of getting from the secure ID to the payment ID. I think this is quite important to call out: if you miss that you need a mapping, it’s unclear how the secure ID works.

source/making_payments/index.html.md.erb
alexbishop1: I think it is but I’ve just checked and a payment can’t be cancelled from the admin tool (only the API).

source/making_payments/index.html.md.erb
alexbishop1: I’m not really sure what this is trying to say.

source/making_payments/index.html.md.erb
alexbishop1: “Malicious users will be unable to guess securely generated UUIDs” works. The only other way a malicious user could access a secure ID is if it’s not stored securely, so maybe there needs to be an emphasis on that.

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,139 @@ weight: 70
...
+</style>
 
-You will receive the ``next_url``  to which you should direct the user to
+<br>
+
+<div class="govuk-warning-text">
+  <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+  <strong class="govuk-warning-text__text">
+    <span class="govuk-warning-text__assistive">Warning</span>
+You must not expose the URL with the ``payment_id`` publicly, for example as a URL parameter or in an insecure cookie. You should store it as a secure cookie or in a database.
```
alexbishop1: This doesn’t quite seem to fit. The `next_url` has a token in it but it’s not the `payment_id`. Bot the `next_url` and the `payment_id` should be kept secure.

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,139 @@ weight: 70
...
+your `return_url` as parameters.
+
+You can match users with payments by using a secure cookie, or creating a
+secure random ID:
+
+### Use a secure cookie
 
-To match up a returning user with their payment, there are two recommended
-methods:
+You can use a secure cookie containing the payment ID from GOV.UK Pay. Your
```
alexbishop1: I think we want to say “encrypted” cookie here. The term “secure cookie” implies that the cookie is sent over HTTPS and has its `secure` flag set but this still allows the user to read and modify it. We need encryption to stop them from doing that.

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,139 @@ weight: 70
...
-your service.
+Users may close their browser or lose internet connection in the middle of
+their payment journey on GOV.UK Pay. In this event, users will not be
+redirected back to your service.
 
-You can still check on the status of these payments by making a GET request
-using the Location Header or Self Link, the same way you would if they were
-redirected, but just after a set time (eg, an hour).
+You can still [check on the status of these payments](#making-payments), but
+after a set time (eg, an hour).
```
alexbishop1: I think an hour is too short because it’s less than the 90 minute expiry timeout we have.

source/making_payments/index.html.md.erb
```diff
@@ -5,120 +5,139 @@ weight: 70
...
+expect occasional notifications of success or failure if users experienced
+problems right at the moment of the redirect.
 
-If a user does not have enough funds in their account to make a payment, the
-current GOV.UK Pay frontend will not let them try again with separate card
-details. They will have to return to the service to retry the payment.
+If a user does not have enough funds in their account to make a payment,
+GOV.UK Pay will not allow them to attempt the payment again with alternative
+card details. Instead, in this situation the user would need to return to the
+service to retry the payment.
```
alexbishop1: This applies more generally than just not having enough money: it’s the same if the card is declined for any reason (e.g. incorrect address or expiry date).

bahmady:
@alexbishop1 I've tried to attend to each of the comments you left: when you have the bandwidth please let me know your thoughts

source/making_payments/index.html.md.erb
bahmady: I've deleted it

source/making_payments/index.html.md.erb
```diff
@@ -5,68 +5,91 @@ weight: 70
...
-You can track the progress of a payment while the user is on GOV.UK Pay using
-the <a
-href="https://gds-payments.gelato.io/api-explorer/gov-uk-pay-api/versions/1.0.2/v1/find-payment-by-id"
-target="blank">Find payment by ID</a> call [external link].
+## Track the progress of a payment
 
-NOTE: The status of the payment will go through several phases until it either
-succeeds or fails. See the [API reference
-section](/api_reference/#api-reference) for more details.
+You can track the progress of a payment while a user is on GOV.UK Pay using
```
alexbishop1: Not sure about the “while a user is on GOV.UK Pay” bit: the payment can be tracked after that and its status may changed after the user has finished with it (capture etc.).

source/making_payments/index.html.md.erb
```diff
@@ -5,68 +5,91 @@ weight: 70
...
 
-* create a secure random ID (such as a UUID) and include this as part of the
-``return_url``, using a different ``return_url`` for each payment. Since a
-securely generated UUID is not guessable, fraudsters will not be able to
-intercept users’ payments.
+### Use an encrypted cookie
 
->Note: If you create an ID yourself, you’ll likely need to store this in a
-datastore mapped to the payment ID just after you create a payment.
+You can use a secure cookie containing the `paymentId` from GOV.UK Pay. Your
```
alexbishop1: This should say “encrypted cookie” rather than “secure cookie”.

source/making_payments/index.html.md.erb
```diff
@@ -77,37 +100,36 @@ datastore mapped to the payment ID just after you create a payment.
  * not paid because of a technical error
```
alexbishop1: There’s also a couple of other reasons:

- their payment being declined (not really a technical error)
- the user cancelled the payment

source/making_payments/index.html.md.erb
```diff
@@ -77,37 +100,36 @@ datastore mapped to the payment ID just after you create a payment.
...
 ## When a user does not complete their payment journey
 
-The user may close their browser or lose internet connection in the middle of
-the payment flow on GOV.UK Pay. These users will not be redirected back to
-your service.
-
-You can still check on the status of these payments by making a GET request
-using the Location Header or Self Link, the same way you would if they were
-redirected, but just after a set time (eg, two hours).
+Users may close their browser or lose internet connection in the middle of
```
alexbishop1: This either wants to be “lose _their_ internet connection” or “lose internet connectivity”.

source/making_payments/index.html.md.erb
```diff
@@ -77,37 +100,36 @@ datastore mapped to the payment ID just after you create a payment.
...
+their payment journey on GOV.UK Pay. In this event, users will not be
+redirected back to your service.
 
->Note: GOV.UK Pay will expire incomplete payments after 90 minutes, but you should expect an occasional success or failure if the user experienced problems right at the moment of the redirect.
+You can still [check the status of these
+payments](/payment_flow/#check-the-status-of-a-payment), but GOV.UK Pay will
+expire incomplete payments after 90 minutes so you should only check after
+this amount of time.
```
alexbishop1: It’s fine for them to check the payment status before the expiry time but they need to be aware that it’s not necessarily final until it’s either completed or expired. Maybe something like this?

“You can still [check the status of these payments](/payment_flow/#check-the-status-of-a-payment), but be aware that GOV.UK Pay allows 90 minutes for a user to complete their payment before expiring it, so you can only make a final decision about whether or not a payment was successful after this time.”

source/making_payments/index.html.md.erb
```diff
@@ -77,37 +100,36 @@ datastore mapped to the payment ID just after you create a payment.
...
 
-If a user does not have enough funds in their account to make a payment, the
-current GOV.UK Pay frontend will not let them try again with separate card
-details. They will be redirected to your service but you can offer them the option to try again. 
-That option would require your service to start a new payment. 
+If a [payment fails](/payment_flow/#payment-failure), GOV.UK Pay will not
+allow the user to attempt the payment again with alternative card details. In
+this situation the user would need to return to the service to retry the
+payment.
```
alexbishop1: Maybe this could be phrased in a way that focuses more on what the service should do to handle these cases. The second sentence could suggest that if the payment fails, the service should offer a way for the user to try again (which creates a new payment).

source/making_payments/index.html.md.erb
```diff
@@ -121,8 +143,8 @@ Use a GET request against a single payment to find out if you can cancel it:
 
 `GET /v1/payments/paymentId`
 
-If the JSON response body contains a `"cancel"` field in the `_links` object that is not `null`, you
-can cancel the payment. For example:
+If the JSON response body contains a `"cancel"` field in the `_links` object
+that is not `null`, you can cancel the payment. For example:
```
alexbishop1: Not entirely clear here that it’s the `"cancel"` field that’s not `null`, rather than the `_links` object.

bahmady: Tried to make clearer by putting in parentheses

source/making_payments/index.html.md.erb
bahmady: Tried to make clearer by adding an extra sentence "Your service should account for this".

source/making_payments/index.html.md.erb
bahmady: Tried to alter language

source/making_payments/index.html.md.erb
bahmady: Indeed, connectivity

source/making_payments/index.html.md.erb
bahmady: Are these distinct from the matter covered in this bullet(?):

> * not paid because their card was rejected or they selected cancel

source/making_payments/index.html.md.erb
bahmady: Indeed, changed

source/making_payments/index.html.md.erb
bahmady: Removed the 'while a user' part

source/making_payments/index.html.md.erb
```diff
@@ -121,8 +150,8 @@ Use a GET request against a single payment to find out if you can cancel it:
 
 `GET /v1/payments/paymentId`
 
-If the JSON response body contains a `"cancel"` field in the `_links` object that is not `null`, you
-can cancel the payment. For example:
+If the JSON response body contains a `"cancel"` field in the `_links` object
+(that is not `null`), you can cancel the payment. For example:
```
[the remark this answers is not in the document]
bahmady: I actually put the parentheses around the wrong part. Should be fixed now.

bahmady:
No worries @alexbishop1 think I've covered all points but feel free to comment again if needed. Thanks again

source/making_payments/index.html.md.erb
```diff
@@ -5,109 +5,137 @@ weight: 70
...
-using the Location Header or Self Link, the same way you would if they were
-redirected, but just after a set time (eg, two hours).
+Users may close their browser or lose internet connectivity in the middle of
+their payment journey on GOV.UK Pay. In this event, users will not be
+redirected back to your service.
 
->Note: GOV.UK Pay will expire incomplete payments after 90 minutes, but you should expect an occasional success or failure if the user experienced problems right at the moment of the redirect.
+You can still [check the status of these
+payments](/payment_flow/#check-the-status-of-a-payment), but GOV.UK Pay will
+expire payments after 90 minutes. You can only be certain of the payment’s
```
alexbishop1: Hmm, I preferred it before.

Maybe it come something like this: “You can still [check the status of these payments](/payment_flow/#check-the-status-of-a-payment), but be aware that GOV.UK Pay will allow a user 90 minutes to complete their payment before it expires. You can only be certain of the payment’s final status after this time.”

source/making_payments/index.html.md.erb
```diff
@@ -121,8 +149,8 @@ Use a GET request against a single payment to find out if you can cancel it:
 
 `GET /v1/payments/paymentId`
 
-If the JSON response body contains a `"cancel"` field in the `_links` object that is not `null`, you
-can cancel the payment. For example:
+If the JSON response body contains a `"cancel"` field (in the `_links` object)
+that's not `null`, you can cancel the payment. For example:
```
alexbishop1: Should be a ’ not a '

PR #225, opened 2018-12-07
Adds integration guide to public documentation

bahmady:
### Context
We lack a 'How to' guide on integrating with GOV.UK Pay

Adds @heathd's doc on integration (minor edits)

### Guidance to review
This is a draft (for publication in the public docs). Some of the images have watermarks for Websequencediagrams.com and I couldn't export them to "high quality PNG" or SVG but Conor has reviewed and is happy that they be published in this PR. 

I've put up a test version of this:

https://govuk-pay-integration.cloudapps.digital/integrate_with_govuk_pay/


source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+### Using a mop-up job
+
+This section describes the recommended approach for most integrations. It
+provides the best guarantee that edge cases are handled properly and
+automatically.
+
+In order to use a mop-up job, you should have a datastore which keeps track of
+incomplete payment journeys.
+
+The following diagram gives a detailed overview of the sequence:
```
katiebates: could be clearer that the following diagram shows the complete payment journey flow

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+determine the status of the payment.
+
+### Integrations without mop-up jobs
+
+You may have a very low-volume service where you have a manual process for
+checking that payments succeeded. For example, using the GOV.UK Pay admin
+tool. You may therefore choose not to use a mop-up job. In this case, you
+should make sure your service staff are trained to check the outcomes of
+payments in the GOV.UK Pay admin tool before releasing the service to the end
+user.
```
katiebates: I have a very slight concern about the use of 'trained' here - since it could make it sound like the process is more complicated than it really is. Could 'you should make sure your service staff check the outcomes of payments..' work instead?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+
+<br>
+
+<div class="govuk-warning-text">
+  <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+  <strong class="govuk-warning-text__text">
+    <span class="govuk-warning-text__assistive">Warning</span>
+There are scenarios in which the `return_url` will never be visited, so there
+may be a small number of payment journeys left in a limbo state. You would
+need to manually reconcile these.
```
katiebates: i'm a bit confused by this warning text 
1. does it apply to payments in the admin tool or API? 
2. can 'limbo state' be defined as to what it actually is (e.g. payments left as 'in progress' or 'created'/'started'/'submitted' - depending on whether the payment is being viewed in the admin tool or API)
if admin tool, then 'in progress' payments are transitioned to 'timed out' after 90 minutes. if a service is relying on the API, then indeed there wouldn't have been a status update.
3. what exactly do we mean by manually reconcile here? manually update the 'service backend'? why would it be manual?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+#### Encode your payment reference in the `return_url` (not recommended)
+
+For example:
+
+`return_url: https://my.service.gov.uk/return/payment_12345`
+
+The risk with this approach is that people may be able to guess the
+`paymentId` in your `return_url`. If you display user information on your
+confirmation screen, you risk exposing users’ personally identifiable
+information to people who should not have access to it.
```
katiebates: How?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+There are scenarios in which the `return_url` will never be visited, so there
+may be a small number of payment journeys left in a limbo state. You would
+need to manually reconcile these.
+  </strong>
+</div>
+
+The advantage of this approach is that you do not need to use a datastore in
+your service.
+
+If you do not use a mop-up job, you have 2 options to identify users when they
```
katiebates: I'm a bit confused by this section, especially in relation to the one above. It suggests that a mop-up job has an additional purpose of identifying users when they return to the return_URL, which I don't think is outlined in the section on mop-up jobs above. In fact, i thought the purpose of the mop-up job was to deal with scenarios where the user **doesn't** return to the return_URL??

Is it actually for scenarios where a service doesn't do a GET request to query the status of the payment when a user is redirected?

source/integrate_with_govuk_pay/index.html.md.erb
katiebates: the final line of the diagram says 'display confirmation screen (success/fail)' - i'm perhaps getting into pedantry here: in payment failure cases, gov.uk pay will have already shown an appropriate payment unsuccessful screen, so we'd expect the service not to show another fail screen, but to direct the user to the next sensible point in the journey (e.g. retrying the payment).

katiebates: also, there's a line in the sequence diagram "find incomplete journey using journey_id" - is that in the right place? i thought you wouldn't see any incomplete journeys in this flow, because the user doesn't get redirected on incomplete journeys..?

source/integrate_with_govuk_pay/index.html.md.erb
katiebates: We agreed: "In this case, you should make sure your service staff are instructed to check the outcomes of payments in the GOV.UK Pay admin tool before releasing the service to the end user."

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -0,0 +1,170 @@
...
+
+In order to use a mop-up job, you should have a datastore which keeps track of
+incomplete payment journeys.
+
+The following diagram gives a detailed overview of the sequence:
+
+![](images/keep-track-of-incomplete-payment-journeys.png)
+
+The following diagram shows an example of an incomplete payment journey, and
+how a mop-up job would clean it up:
```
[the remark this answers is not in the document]
katiebates: diagram: 'times out and marked as failed'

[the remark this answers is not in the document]
katiebates: >3 hours gives a margin of error

source/integrate_with_govuk_pay/index.html.md.erb
katiebates: 1. 'The following diagram gives a detailed overview of the payment flow including use of a datastore' 
2. 'display appropriate screen' _might_ be better 
3. ignore my comment

bahmady:
Sent off for writer (2i) review; ETA early next week

PR #226, opened 2018-12-11
Calls out need to sign up for API explorer in QS guide

bahmady:
### Context
An easy addition to call out the need to sign up for the API explorer as part of the QS guide

Adds a line and a link to Gelato

### DIFF
--- a/source/quick_start_guide/index.html.md.erb
+++ b/source/quick_start_guide/index.html.md.erb
@@ -5,8 +5,6 @@ weight: 40
 
 # Quick start guide
 
-<br>
-
 <style>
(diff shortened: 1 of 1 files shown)


source/quick_start_guide/index.html.md.erb
```diff
@@ -29,12 +27,15 @@ Before you start using GOV.UK Pay, you should:
...
 Pay is right for your service
 
-* sign up for [the GOV.UK Pay admin
+* sign up to the [GOV.UK Pay admin
 site](https://selfservice.payments.service.gov.uk/create-service/register)
 
 * read the GOV.UK Service Manual guidance on [deploying new
 software](https://www.gov.uk/service-manual/making-software/deployment.html)
 
+* sign up to the [API
```
sfount: I think making it clear that you will have to 'sign up' makes this clearer 👍 . For GitHub devs it's just a one click sign in but for others they will have to go through a full account creation process.

It is possible to use gelato without signing in you would just have to add your own `Authorizaion` header, maybe explaining this complicates the flow too much?

bahmady: Yeah I did consider devs might not have a GitHub account. Not sure about the note with the `Authorization` header as I agree it could be a little bit too much, but we're going to be changing the entire flow in the new year when we replace Gelato in any case.

PR #236, opened 2018-12-21
Fixes broken Gelato links

bahmady:
Fixes 'Search payments' broken links to Gelato (API explorer)


bahmady:
In hindsight you're right; I've amended

PR #239, opened 2018-12-27
Added that HTTP can be used in test accounts but not in live accounts

tillwirth:


bahmady:
@tillwirth I removed a negative contraction and the reference to "we" outside a tutorial context, to make this to style. I also added a link to the relevant section in the docs re. using HTTPS for return URLs with test accounts. 

Otherwise this looks good, thank you. If you're happy I'll merge

PR #251, opened 2018-12-28
Adds Maestro

bahmady:
### Context
@tillwirth pointed out via email that we lack 'Maestro' in https://docs.payments.service.gov.uk/api_reference/#card-types

Adds Maestro to list of cards in API reference section

### Guidance to review
This isn't my area of expertise so I don't know that my proposed edit at time of writing is correct, so I'm opening this as a pointer that this change is needed as much as anything

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -232,6 +232,7 @@ For example:
 | visa             | CREDIT | Visa             |
 | master-card      | DEBIT  | Mastercard       |
 | master-card      | CREDIT | Mastercard       |
+| maestro          | DEBIT  | Maestro          |
 | american-express | CREDIT | American Express |
 | diners-club      | CREDIT | Diners Club      |
 | discover         | CREDIT | Discover         |


source/api_reference/index.html.md.erb
```diff
@@ -232,6 +232,8 @@ For example:
 | visa             | CREDIT | Visa             |
 | master-card      | DEBIT  | Mastercard       |
 | master-card      | CREDIT | Mastercard       |
+| maestro          | DEBIT  | Maestro          |
+| maestro          | CREDIT | Maestro          |
```
rhowe-gds: I don't think there is such a thing as a Maestro credit card

rhowe-gds: Certainly there isn't anything in the BIN ranges for one

bahmady: I didn't think that either but Wikipedia has confused me: https://en.wikipedia.org/wiki/Maestro_(debit_card)

> Maestro (stylized as maestro) is a multi-national credit card service ...

tillwirth: @rhowe-gds is right, we need to remove Maestro Credit

bahmady: No problem, done. Thanks both 👍

PR #252, opened 2018-12-28
Amend description re. payment status failure in admin tool 

bahmady:
### Context
@tillwirth wrote via email:

> The following section/statement is no longer accurate: 
> https://docs.payments.service.gov.uk/api_reference/#use-the-admin-tool
> "Select a payment with a "failed", "cancelled" or "error" status."
> We have replaced the "failed" state with "new" states that only exist in the admin tool but not the  API: Declined and Timed Out. We may have also combined the "user cancelled" and "system cancelled" into one "cancelled" state in the admin tool but I'm not sure about that. 

I confirmed that the "cancelled" statuses are indeed consolidated in the self-service tool 

Suggested edit to account for the above

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -201,8 +201,8 @@ Pay admin tool or directly using the API.
 
 1. Sign in to the [GOV.UK Pay admin
 site](https://selfservice.payments.service.gov.uk/) (link opens in new window).
-2. Select _Transactions_.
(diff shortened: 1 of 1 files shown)


bahmady:
@tillwirth you may have seen on Slack that @katiebates and I discussed putting a diagram together that maps API states to admin tool ones, and placing it e.g. here: https://docs.payments.service.gov.uk/api_reference/#use-the-admin-tool

If we're happy with this particular PR I'll merge

PR #257, opened 2019-01-08
State that the amount is a number

jonathanglassman:
### Context

HMPO feedback on the documentation states that it would be useful to say that a Number is required for the parameter "amount" when Posting to the /payments endpoint.

Add statement to the Payment flow overview section that the amount parameter is a number data type.

### Guidance to review

This is a small change based on HMPO feedback and just needs review and approval from a Pay approved reviewer.

### DIFF
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -102,7 +102,7 @@ For example:
 }
 ```
 
-`amount`: The amount in pence - in the example, the payment is for £145.
+`amount`: The amount in pence - in the example, the payment is for £145. This value must be a number data type.
 
 `reference`: The reference number you wish to associate with this payment. The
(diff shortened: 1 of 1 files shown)


source/payment_flow/index.html.md.erb
```diff
@@ -102,7 +102,7 @@ For example:
 }
 ```
 
-`amount`: The amount in pence - in the example, the payment is for £145.
+`amount`: The amount in pence - in the example, the payment is for £145. This value must be a number data type.
```
alexbishop1: I wonder if explicitly putting that it should be a number and not a string might be useful (since that’s the most common way people mess this up)

jonathanglassman: If I change are you happy for it to be merged?

alexbishop1: Sure

PR #258, opened 2019-01-11
Change docs to reflect new Slate instance rather than Gelato

bahmady:
### Context
We're moving away from Gelato, to Slate, for the "endpoint" part of our API documentation offering. We need to remove references to it, and to an "API explorer" generally. We also need to change some surrounding text, where relevant, particularly in the Quick start guide. The Slate instance is here:

https://govukpay-api-browser.cloudapps.digital/

@jonathanglassman and I have been through and changed the docs to satisfy the above. We refer to the new Slate instance as the "API browser". 

### Guidance to review
This should be reviewed by others because it involves a lot of changes and we need to make sure they make sense, and that the links work. @jonathanglassman and I have reviewed together and we're happy that it's OK, but it's important to get this right. Ideally @heathd or @alexbishop1 will review to confirm that from a tech perspective it's all good. 

Live preview of docs changes: https://govukpay-api-browser-changes.cloudapps.digital/

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -14,19 +14,13 @@ determines if actions are treated as test payments, or processed as real payment
 The API is based on REST principles. It returns data in JSON format, and uses
 standard HTTP error response codes.
 
-For full details of each API action, see the API browser:
-
-* <a href="https://gds-payments.gelato.io/docs/versions/1.0.2/resources/general" target="blank">General functions</a> [external link]
-
-* <a href="https://gds-payments.gelato.io/docs/versions/1.0.2/resources/payment-id" target="blank">Payment ID functions</a> [external link]
-
-You can also use the interactive <a
-href="https://gds-payments.gelato.io/api-explorer/gov-uk-pay-api/versions/1.0.2//"
-target="blank">API explorer</a> [external link] to try API calls and view responses.
+For full details of each API action, see the <a
+href="https://govukpay-api-browser.cloudapps.digital/" target="blank">API
+browser</a> [external link].
 
 See the [__Quick start guide__](/quick_start_guide/#quick-start-guide) section
-to read how to set up the API explorer. Make sure you enter your test API
-key to avoid generating real payments.
+to read how to get started with the API quickly. Make sure you enter your test
+API key to avoid generating real payments.
 
 ## Authentication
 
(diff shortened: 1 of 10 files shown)


bahmady:
Live preview of docs changes: https://govukpay-api-browser-changes.cloudapps.digital/

katiebates:
i've tested this by searching 'external link' in the search box. 

There's one instance where it's not linked: https://govukpay-api-browser-changes.cloudapps.digital/payment_flow/#making-a-payment in this section there's an [external link] but it doesn't link anywhere 

Otherwise it looks like all instances of Gelato have been replaced by Slate, and it looks like all the external links take the user to the appropriate part of the API browser. 

My only other comment is whether we should mention that users can use tools like Postman (etc) when making a test API call in this section? 
https://govukpay-api-browser-changes.cloudapps.digital/quick_start_guide/#make-a-test-api-call

jonathanglassman:
> i've tested this by searching 'external link' in the search box.
> 
> There's one instance where it's not linked: https://govukpay-api-browser-changes.cloudapps.digital/payment_flow/#making-a-payment in this section there's an [external link] but it doesn't link anywhere
> 
> Otherwise it looks like all instances of Gelato have been replaced by Slate, and it looks like all the external links take the user to the appropriate part of the API browser.
> 
> My only other comment is whether we should mention that users can use tools like Postman (etc) when making a test API call in this section?
> https://govukpay-api-browser-changes.cloudapps.digital/quick_start_guide/#make-a-test-api-call

Hi @katiebates thanks for the review. I cannot see the non-linked content you are referring to in the payment flow section. There are two links that i can see, one to https://govukpay-api-browser.cloudapps.digital/#newpayment and one to https://govukpay-api-browser.cloudapps.digital/#getpayment, and both work. Am I completely missing something?

Regarding the tools question, as long as we mention multiple tools so we don't look like we are favouring any single one, that should be fine, I'll add that content in now.

katiebates:
sorry my bad, there is a link. it's just that the [external link] is a few words after the hyperlink so i didn't spot it immediately :)

jonathanglassman:
> sorry my bad, there is a link. it's just that the [external link] is a few words after the hyperlink so i didn't spot it immediately :)

No problem and thanks for the review. I've added in the API testing tool examples. Would appreciate a look from @kbottla or @heathd to ensure I've used appropriate examples.

jonathanglassman:
@alexbishop1 change made

PR #265, opened 2019-01-31
PP-4764 Make Worldpay link clickable

DanailMinchev:
### Context

Add whitespace `&nbsp;` between the `Worldpay` link and the title `The differences between test and live accounts`, because the CSS of the title makes the link un-clickable.
Avoided CSS changes not to break other pages in terms of displaying titles.

with @jonheslop

### DIFF
--- a/source/stylesheets/screen-old-ie.css.scss
+++ b/source/stylesheets/screen-old-ie.css.scss
@@ -2,3 +2,11 @@ $is-ie: true;
 $ie-version: 8;
 
 @import "govuk_tech_docs";
+
+.has-sidebar .technical-documentation {
+  h1,
+  h2,
+  h3 {
+    margin-top: 0px !important;
+  }
+}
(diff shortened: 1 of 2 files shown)


jonheslop:
Hmm thats annoying a better fix would be to remove the negative margin off the headings so the link doesnt overlap the things above it  
screen shot 2019-01-31 at 16 38 05

jonathanglassman:
@jonheslop would removing the negative margin be a quick fix? If so then we should implement for the tech doc tool and push an update. If it is not quick, then adding in the line break works even if it isn't perfect.

DanailMinchev:
Thanks @jonheslop 
@jonathanglassman we pushed the changes, let me know if this can be merged or we want to merge it in the doc tool, thank you

jonathanglassman:
@DanailMinchev @jonheslop thanks both. Can we put this fix in the tech doc tool so we can roll it out to all docs?

selfthinker:
This has now been properly fixed upstream: https://github.com/alphagov/tech-docs-gem/pull/74

jonheslop:
Thanks will revert this and update our version

PR #273, opened 2019-02-19
Create digital wallets content

jonathanglassman:
### Context

Google Pay and Apple Pay are launching.

New content for Apple Pay and Google Pay.

### DIFF
--- a/source/optional_features/digital_wallets/index.html.md.erb
+++ b/source/optional_features/digital_wallets/index.html.md.erb
@@ -7,7 +7,7 @@ weight: 136
 
 You can enable [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
 
-This is an invitation-only feature. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service.
+This is an invitation-only feature. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service. After GOV.UK Pay has approved your request to participate in the private beta for Apple Pay and Google Pay, you will receive an email with a link to enable them. 
 
 You can only enable Google Pay and Apple Pay if you have a [live Worldpay account](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account).
 
@@ -15,9 +15,7 @@ You can only enable Google Pay and Apple Pay if you have a [live Worldpay accoun
 
 Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
 
-1. Go to __Settings__ and select __Card types__.
-
-1. Select the __Digital wallet__ tab.
+1. Click on the link in the email from GOV.UK Pay 
 
 1. Select __Enable__ next to Apple Pay.
 
@@ -41,9 +39,7 @@ You get your Google Pay merchant ID from your Worldpay merchant admin interface.
 
 Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
 
-1. Go to __Settings__ and select __Card types__.
-
-1. Select the __Digital wallet__ tab.
+1. Click on the link in the email from GOV.UK Pay 
 
 1. Select __Enable__ next to Google Pay.
 
@@ -66,7 +62,7 @@ Google Pay does not have biometric authentication enabled by default.
 
 The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
 
-When you look at individual non-digital wallet transaction in your GOV.UK Pay account transactions list, you see the six digits and last four digits of the card number. When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number or device account number.
+When you look at individual non-digital wallet transaction in your GOV.UK Pay account transactions list, you see the first six digits and last four digits of the card number. When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number or device account number.
 
 ### Google Pay
 
--- a/source/optional_features/digital_wallets/index.html.md.erb
+++ b/source/optional_features/digital_wallets/index.html.md.erb
@@ -7,6 +7,8 @@ weight: 136
 
 You can enable [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
 
+This is an invitation-only feature. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service.
+
 You can only enable Google Pay and Apple Pay if you have a [live Worldpay account](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account).
 
 ## Enable Apple Pay
@@ -51,18 +53,21 @@ Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.
 
 ## 3D Secure
 
-If a digital wallet transaction requires 3D Secure authentication, and the user making that transaction has not enabled biometric authentication on their digital wallet, that user must complete 3D Secure authentication in the same way as a non-digital wallet transaction.
+A user making a digital wallet transaction must complete 3D secure authentication on that transaction if:
 
-Apple Pay has biometric authentication enabled by default, so Apple Pay transactions should not require 3D Secure authentication.
+* that transaction needs 3D Secure authentication
+* that user has not enabled biometric authentication on their digital wallet
 
-Google Pay does not have biometric authentication enabled by default. If your users have not enabled biometric authentication on their Google Pay digital wallet and they make any transactions that require 3D Secure authentication, those users must complete 3D Secure authentication in the same way as non-digital wallet transactions.
+Apple Pay has biometric authentication enabled by default, so users should not have to complete 3D Secure authentication during Apple Pay transactions.
+
+Google Pay does not have biometric authentication enabled by default. 
 
 ## Restrictions
 
 The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
 
-When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number.
+When you look at individual non-digital wallet transaction in your GOV.UK Pay account transactions list, you see the six digits and last four digits of the card number. When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number or device account number.
 
 ### Google Pay
 
-You cannot restrict Google Pay digital wallet transactions by card type. For example, you cannot only accept digital wallet transactions made with a debit card.
+You can restrict Google Pay digital wallet transactions by card scheme, but not by card type. For example, you can state that you only accept digital wallet transactions made with Visa cards, but you cannot only accept these transactions made with a debit card.
--- a/source/optional_features/digital_wallets/index.html.md.erb
+++ b/source/optional_features/digital_wallets/index.html.md.erb
@@ -5,9 +5,9 @@ weight: 136
 
 # Digital wallets
 
-You can enable [Google Pay](https://pay.google.com) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
+You can enable [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
 
-You can only enable Google Pay and Apple Pay if you use [Worldpay](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account) as your payment service provider (PSP).
+You can only enable Google Pay and Apple Pay if you have a [live Worldpay account](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account) as your payment service provider (PSP).
 
 If you use Stripe, [ePDQ](/switching_to_live/set_up_a_live_epdq_account/#set-up-a-live-epdq-account) or [Smartpay](/switching_to_live/set_up_a_live_smartpay_account/#set-up-a-live-smartpay-account) as your PSP, you cannot enable Apple Pay or Google Pay.
 
@@ -59,6 +59,12 @@ If a digital wallet transaction requires 3D Secure authentication, your users mu
 
 The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
 
-You cannot restrict digital wallet transactions by card type. For example, you cannot only accept digital wallet transactions made with a debit card.
-
 When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number.
+
+The expiry date for individual digital wallet transactions in your GOV.UK Pay account transactions list is the expiry date of the XYZ certificate rather than the payment expiry date.
+
+### Google Pay
+
+You cannot restrict Google Pay digital wallet transactions by card type. For example, you cannot only accept digital wallet transactions made with a debit card.
+
+When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the cardholder name if you have enabled [billing address collection](/integrate_with_govuk_pay/#collecting-your-users-39-billing-addresses).
--- a/source/optional_features/digital_wallets/index.html.md.erb
+++ b/source/optional_features/digital_wallets/index.html.md.erb
@@ -7,7 +7,7 @@ weight: 136
 
 You can enable [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
 
-This is an invitation-only feature currently in private beta. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service. After GOV.UK Pay has approved your request to participate in this private beta, GOV.UK Pay will email you a link to enable this feature. 
+This is an invitation-only feature currently in private beta. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service. After GOV.UK Pay has approved your request, GOV.UK Pay will email you a link to enable this feature. 
 
 You can only enable Google Pay and Apple Pay if you have a [live Worldpay account](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account).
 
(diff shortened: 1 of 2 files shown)


source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,64 @@
+---
+title: Digital wallets
+weight: 136
+---
+
+# Digital wallets
+
+You can enable [Google Pay](https://pay.google.com) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
+
+You can only enable Google Pay and Apple Pay if you use [Worldpay](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account) as your payment service provider (PSP).
```
gidsg: This will not work in sandbox nor should you be able to enable these in sandbox. So I think it'd be better to say you must have a *live WorldPay account* or words to that effect.

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,64 @@
...
+
+If you use Stripe, [ePDQ](/switching_to_live/set_up_a_live_epdq_account/#set-up-a-live-epdq-account) or [Smartpay](/switching_to_live/set_up_a_live_smartpay_account/#set-up-a-live-smartpay-account) as your PSP, you cannot enable Apple Pay or Google Pay.
```
katiebates: although we could say that Apple/Google Pay will be supported for Stripe soon?

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,64 @@
...
+
+## 3D Secure
+
+If a digital wallet transaction requires 3D Secure authentication, your users must complete 3D Secure authentication in the same way as non-digital wallet transactions.
+
+## Restrictions
+
+The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
+
+You cannot restrict digital wallet transactions by card type. For example, you cannot only accept digital wallet transactions made with a debit card.
```
katiebates: this only applies to Google Pay

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,64 @@
...
+
+When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number.
```
katiebates: i think you also don't see the actual card expiry date - but Oz or @danworth can elaborate 

with google pay, you only receive cardholder name if billing address collection is enabled (i think)

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: agree

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,64 @@
+---
+title: Digital wallets
+weight: 136
+---
+
+# Digital wallets
+
+You can enable [Google Pay](https://pay.google.com) and [Apple Pay](https://www.apple.com/uk/apple-pay/) [external links] to take payments from your users.
```
alexbishop1: If you’re not logged in to a Google account, https://pay.google.com/ takes you to a login page. I think a URL like https://pay.google.com/intl/en_uk/about/ would be better (this does not require you to be logged in to Google).

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: Do we know when? Best practice for tech docs is to avoid using imprecise time terms as they lead to more questions / uncertainty. We should either say for example "You will be able to use Apple Pay and Google Pay with Stripe in TIME PERIOD e.g. summer 2019" or leave as is and update when it changes. I recommend keeping as is, and using the suggested content in an announcement or communication to Pay users about the new functionality.

source/optional_features/digital_wallets/index.html.md.erb
alexbishop1: At least for Apple Pay (not sure about Google Pay), the expiry date is a date but it’s not the card expiry date (it’s some certificate expiry date the user won’t know anything about).

Katie’s correct that Google Pay will only give us the cardholder name if billing address collection is enabled (it’s always available with Apple Pay).

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: will clarify

source/optional_features/index.html.md.erb
```diff
@@ -7,12 +7,14 @@ weight: 130
 
 GOV.UK Pay supports some optional features which need configuration:
 
-* [Custom branding](/optional_features/custom_branding/#custom-branding-of-payment-pages) of payment pages
+* [custom branding](/optional_features/custom_branding/#custom-branding-of-payment-pages) of payment pages
 
-* [Welsh language](/optional_features/welsh_language/#welsh-language-payment-pages) of payment pages
+* [welsh language](/optional_features/welsh_language/#welsh-language-payment-pages) of payment pages
```
alexbishop1: “Welsh” is a proper noun 😀

jonathanglassman: oops!

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: will change

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: Do we know which certificate expiry date it is?

source/optional_features/digital_wallets/index.html.md.erb
oswaldquek: do we even need line 12, as 11 explicitly says Google/Apply Pay can only be enabled if the PSP is Worldpay.

source/optional_features/digital_wallets/index.html.md.erb
oswaldquek: yes one can't see the card expiry date with Google Pay

oswaldquek: i don't even know what this certificate expiry is.... is it even useful for our clients?

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,70 @@
...
+
+The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
+
+When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number.
+
+The expiry date for individual digital wallet transactions in your GOV.UK Pay account transactions list is the expiry date of the XYZ certificate rather than the payment expiry date.
+
+### Google Pay
+
+You cannot restrict Google Pay digital wallet transactions by card type. For example, you cannot only accept digital wallet transactions made with a debit card.
```
oswaldquek: i think this is true for Apple Pay as well @katiebates @jonheslop ?

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,70 @@
...
+
+When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the cardholder name if you have enabled [billing address collection](/integrate_with_govuk_pay/#collecting-your-users-39-billing-addresses).
```
oswaldquek: This isn't true. We collect the name and email address of the user when the user makes a card payment. So the name will appear on the transactions list even if the billing address wasn't collected.

oswaldquek: Although, strictly speaking, the user can enter a dummy name, so it's not necessarily the real cardholder name.

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: That's why I wanted to check which certificate it is. If it isn't something they would normally need to know, then we can probably not mention it. But can't really judge that if we aren't sure what the certificate is.

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: It's a fair point, we could remove. @katiebates what do you think?

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: Just to build consensus, is everyone happy to remove this? @katiebates @alexbishop1 ?

source/optional_features/digital_wallets/index.html.md.erb
alexbishop1: @jonheslop might know which certificate it is

source/optional_features/digital_wallets/index.html.md.erb
katiebates: fine by me

source/optional_features/digital_wallets/index.html.md.erb
alexbishop1: It depends on how likely it is that the name will be different to the cardholder name and whether this matters. Where does the name come from? Is it the name associated with the Google account?

I can think of a few reasons why the cardholder name might differ from the Google account name:
- Minor differences like the cardholder name being “Mr Alexander Bishop” and the Google account name being “Alex Bishop”
- The Google account name being completely different to the cardholder name e.g. an alias — I don’t know how hard Google tries to persuade people to provide their real name
- The Google account being in the name of one person but the cardholder name being in the name of another person — I can imagine this might be common for families (Google account in the name of one family member, card in the name of another)

The most important question is whether services will actually care about any differences?

source/optional_features/digital_wallets/index.html.md.erb
alexbishop1: Nah, Apple Pay allows distinguishing between debit and credit cards:

https://developer.apple.com/documentation/apple_pay_on_the_web/applepaypaymentrequest/1916123-merchantcapabilities

I assume we’ve hooked this up to our existing settings for accepting debit and/or credit cards.

jonathanglassman: ok will leave as is

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: I have no idea. I think that is a product question. From a content perspective, I don't think it makes a difference whether it is included or not.

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,70 @@
...
+
+1. Select __Enable__ next to Google Pay.
+
+1. On the __Enable Google Pay__ screen, select __Yes, turn on Google Pay__.
+
+1. Enter the generated merchant ID and select __Submit__.
+
+## 3D Secure
+
+If a digital wallet transaction requires 3D Secure authentication, your users must complete 3D Secure authentication in the same way as non-digital wallet transactions.
```
katiebates: could this section be a little more specific? eg 3DS isn't used on any Apple journeys and only on some google pay journeys, so it's a better user experience, higher conversion etc however you'll sometimes see it with google pay where biometric auth isn't used.

jonathanglassman: How about: 

"If a digital wallet transaction requires 3D Secure authentication and that digital wallet does not have biometric authentication enabled, your users must complete 3D Secure authentication in the same way as non-digital wallet transactions.

Apple Pay has biometric authentication enabled by default so Apple Pay transactions should not require 3D secure authentication. 

Google Pay does not have biometric authentication enabled by default. If your users do not have biometric authentication on their Google Pay digital wallet and they make any transactions that require 3D Secure authentication, those users must complete 3D Secure authentication in the same way as non-digital wallet transactions."

It's a bit long-winded but relatively clear

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: Removing as per discussion with @katiebates

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: including as per discussion with @katiebates

source/optional_features/digital_wallets/index.html.md.erb
jonathanglassman: removing as per discussion with @katiebates and @jonheslop

oswaldquek:
This looks good to me, but will defer to @katiebates to approve it.

katiebates:
doublechecking with google pay that it's the absence of biometric authentication which means that the payment can fall into scope of 3DS. In a previous reply they'd indicated it was more to do with tokenisation of the card - i.e. if tokenised, 3DS not required.

katiebates:
google replied- "That is correct. Especially on the web, if your PSP asks for a 3D challenge you should put the users through it."

jonathanglassman:
> google replied- "That is correct. Especially on the web, if your PSP asks for a 3D challenge you should put the users through it."

Great, thank you. I'll put it in for 2i now.

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,68 @@
...
+
+1. On the __Enable Google Pay__ screen, select __Yes, turn on Google Pay__.
+
+1. Enter the generated merchant ID and select __Submit__.
+
+## 3D Secure
+
+If a digital wallet transaction requires 3D Secure authentication, and the user making that transaction has not enabled biometric authentication on their digital wallet, that user must complete 3D Secure authentication in the same way as a non-digital wallet transaction.
+
+Apple Pay has biometric authentication enabled by default, so Apple Pay transactions should not require 3D Secure authentication.
```
jonheslop: Could we state that it’s the user that won’t be required to go through 3DS. So like 

```suggestion
Apple Pay has biometric authentication enabled by default, so during Apple Pay transactions users will not be required to go though the usual 3D Secure authentication steps.
```

Realise thats wordy and happy to be wrong?

jonathanglassman: Thanks @jonheslop. I would tweak it slightly; we don't use terms like "the usual" because that can be subjective. Also, I would like to keep "should", as I imagine people can disable biometric auth within their apple pay wallet. How about: 
Apple Pay has biometric authentication enabled by default, so users should not have to go through 3D Secure authentication during Apple Pay transactions.


will not be required to go though the usual 3D Secure authentication steps.

gidsg: @jonathanglassman My understanding from https://support.apple.com/en-gb/HT201239 is that without biometrics the user would be prompted for a passcode and this still counts as a sufficiently secure transaction so there is no circumstances in which Apple Pay will trigger 3DS, so 
> will not 

is a more accurate wording.

jonathanglassman: Thanks @gidsg will redraft

gidsg: @jonathanglassman ...may be more accurate to say *Apple Pay has 2 factor authentication* as this would encompass biometrics or a passcode.

jonathanglassman: @gidsg So, should we in fact say:

```
A user making a digital wallet transaction must complete 3D secure authentication on that transaction if:

* that transaction needs 3D Secure authentication
* that user has not enabled two-factor authentication on their digital wallet

Apple Pay has two-factor authentication enabled by default, so users will not have to complete 3D Secure authentication during Apple Pay transactions.

Google Pay does not have two-factor authentication enabled by default.
```

This covers off both biometrics and other 2fa methods.

alexbishop1: Not sure _2 factor_ works because there isn’t really a first factor!

I think only using a passcode to authenticate only applies to Apple Pay Cash payments in store. On the Web, I think it always requires biometrics (Face ID or Touch ID) on on iPhone/iPad/Mac, or double-clicking the side button on an Apple Watch (I guess the fact that it’s attached to your wrist is proof enough that it’s you!).

I think saying that Apple Pay requires authentication and therefore 3D Secure won’t be used would be fine.

jonathanglassman: So, is the following ok:

```
A user making a digital wallet transaction must complete 3D secure authentication on that transaction if:

* that transaction needs 3D Secure authentication
* that user has not enabled authentication on their digital wallet

Apple Pay enables authentication by default, so users do not have to complete 3D Secure authentication during Apple Pay transactions.

Google Pay does not enable authentication by default, so users may have to complete 3D Secure authentication during Google Pay transactions, depending on their configuration.
```

alexbishop1: Apple Pay doesn’t just enable authentication by default, it enforces it always. You can’t not use authentication with Apple Pay.

Also, rather than talking about the user having authentication enabled on their digital wallet, I think it might be simpler to talk about whether the user used authentication when making the payment (e.g. in the second bullet point, something like “the user did not securely authenticate when using their digital wallet to confirm the payment”).

jonathanglassman: To clarify, "the user did not securely authenticate...the payment" means that there are other methods of secure authentication apart from 3D Secure, and the user did not use any of these non-3D Secure methods, e.g. biometrics

jonathanglassman: @alexbishop1 @oswaldquek @tillwirth @katiebates @gidsg please take a look at the latest draft. I've changed the focus of the authentication section to be more about secure authentication overall, with a reference to 3D Secure as a subset of secure auth.

jonathanglassman:
thank you @tillwirth for approving. Does anyone else have any further feedback? The feature is being shipped with a broken link to the not-live-yet documentation. I will send to 2i on Tuesday 5 March so please do feedback at the start of that day if you can. I aim to get the docs through 2i and live by Friday 8 March.

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,69 @@
...
+
+This is an invitation-only feature currently in private beta. [Contact the GOV.UK Pay team](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) to request this feature for your service. After GOV.UK Pay has approved your request to participate in this private beta, GOV.UK Pay will email you a link to enable this feature. 
+
+You can only enable Google Pay and Apple Pay if you have a [live Worldpay account](/switching_to_live/set_up_a_live_worldpay_account/#set-up-a-live-worldpay-account).
+
+## Enable Apple Pay
+
+Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
+
+1. Click on the link in the email from GOV.UK Pay
```
alexbishop1: This bullet point is missing a period at the end

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,69 @@
...
+
+1. Go to __Integration__ and select the __Pay with Google__ tab.
+
+1. Under __Generate unique merchant ID__, select __Generate__.
+
+1. Copy the generated merchant ID.
+
+Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
+
+1. Click on the link in the email from GOV.UK Pay
```
alexbishop1: This bullet point is missing a period at the end

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,69 @@
...
+## Secure authentication
+
+A user can securely authenticate digital wallet transactions that require secure authentication by either:
+
+* enabling secure authentication in their digital wallet
+* securely authenticating those transactions separately using [3D Secure authentication](https://www.pcisecuritystandards.org/documents/FAQs_for_PCI_3DS_Core_Security_Standard.pdf) [external link]
+
+Apple Pay enforces secure authentication, so users should not have to complete 3D Secure authentication during Apple Pay transactions.
+
+Google Pay does not enforce secure authentication.
```
alexbishop1: Technically this is correct but I’m not too sure about it. We’re introducing this concept of “secure authentication” that’s distinct from 3D Secure authentication but I’m not certain all readers will fully grasp the specificity of “secure authentication” in phrases like “Google Pay does not enforce secure authentication.” (The distinction is actually a bit blurry: when secure authentication is used with a digital wallet, it gets processed using the 3D Secure mechanism even though the user never sees 3D Secure.)

I think the important part for services is whether or not users will see 3D Secure, so I’d anchor the text around that. Something like this (phrasing obviously needs improvement):

“Users paying with Apple Pay never need to complete 3D Secure authentication because Apple Pay requires transactions to be authenticated using a secure mechanism like a fingerprint.

“Google Pay supports authenticating transactions using a fingerprint or similar secure mechanism, but does not require it. Users who do not use Google Pay’s secure authentication may have to complete 3D Secure authentication.”

I think the 3D Secure link is probably too technical for most of our target audience. I mentioned fingerprint authentication as an example because that’s probably something readers are familiar with and can relate to.

source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -0,0 +1,69 @@
...
+
+Apple Pay enforces secure authentication, so users should not have to complete 3D Secure authentication during Apple Pay transactions.
+
+Google Pay does not enforce secure authentication. 
+
+## Restrictions
+
+The GOV.UK Pay platform cannot apply [corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate_card_surcharges) to digital wallet transactions made with a corporate card.
+
+When you look at individual non-digital wallet transaction in your GOV.UK Pay account transactions list, you see the first six digits and last four digits of the card number. When you look at individual digital wallet transactions in your GOV.UK Pay account transactions list, you will only see the last 4 digits of the card number or device account number.
```
alexbishop1: Does “device account number” need defining? I guess we’re basically saying that they’ll see four digits but they might not necessarily match those on the physical card.

Also, inconsistent usage of “four” versus “4”.

jonathanglassman:
Approved by 2i with minor tweaks. Merging.

PR #274, opened 2019-02-21
Update testing with new mock card numbers

jonathanglassman:
### Context

Our mock card numbers content needs clarifying as per some user feedback we have received.

- Adds missing mock card numbers
- Removes unnecessary and confusing links to PSP content
- Clarifies that you must use test card numbers with a test account


jonathanglassman:
Passed 2i with minor tweaks. Will merge and close

PR #286, opened 2019-05-13
Prefill fields on the payment page

m-green:
### Context
When service teams create a payment through the API, they can provide user details that they've already collected in their service. This will then prefill fields on the payment page. 

Add a new page about prefilling fields, and a link from the 'Payment flow' page.

### Guidance to review
Please review for factual accuracy.

### DIFF
--- a/source/optional_features/prefill_user_details/index.html.md.erb
+++ b/source/optional_features/prefill_user_details/index.html.md.erb
@@ -1,22 +1,22 @@
 ---
-title: Prefill fields on the payment page
+title: Prefilling fields on the payment page
 weight: 137
 ---
 
-# Prefill fields on the payment page
+# Prefilling fields on the payment page
 
-If you collect a user's details in your service before you redirect them to GOV.UK Pay, you can prefill some of the fields on the __Enter card details__ page. This means users should not need to enter their details twice.
+If you collect a user's details in your service before you redirect them to GOV.UK Pay, you can prefill some of the fields on the __Enter card details__ page. If you prefill fields, users should not need to enter their details twice.
 
-Users can still edit the details before they complete their payment.
+Users can still edit these details before they complete their payment.
 
-To prefill fields, you can include the following optional parameters when you [create a new payment](/making_payments/#making-payments):
+To prefill fields, include the following optional parameters in the <a href="https://govukpay-api-browser.cloudapps.digital/#newpayment" target="blank">API request</a> when you [create a new payment](/payment_flow/#making-a-payment):
 
 - `email`
 - `prefilled_cardholder_details` - with optional `cardholder_name` and `billing_address` parameters
 
 For example:
 
-```json
+```javascript
 "email": "sherlock.holmes@example.com",
 "prefilled_cardholder_details": {
     "cardholder_name": "Sherlock Holmes",
@@ -29,7 +29,7 @@ For example:
  }
 ```
 
-All of the parameters in `billing_address` are optional.
+All the parameters in `billing_address` are optional.
 
 If you include the `country` parameter, it must be an <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
 
@@ -37,7 +37,7 @@ Refer to the <a
 href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
 target="blank">API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more information.
 
-## Which data is collected
+## Data collected by your GOV.UK Pay admin account
 
 After the user completes their payment, your [GOV.UK Pay admin
 account](https://selfservice.payments.service.gov.uk/login) will collect the user's:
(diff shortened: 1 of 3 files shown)


source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -29,9 +29,9 @@ For example:
  }
 ```
 
-You can leave out any of the parameters.
+All of the parameters in `billing_address` are optional.
 
-The `country` parameter must be a valid <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field will default to 'GB'.
+If you include the `country` parameter, it must be a valid <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field will default to 'Great Britain'.
```
SandorArpa: It actually defaults to country code 'GB' but the user is going to see 'United Kingdom'

m-green: Ah thanks for clarifying @SandorArpa!

source/payment_flow/index.html.md.erb
```diff
@@ -120,7 +120,10 @@ they have completed their payment attempt on GOV.UK Pay. This should not be
 JSON-encoded, because backslashes are invalid characters. With test accounts,
 you can [use HTTP for return URLs instead of HTTPS](/security/#https).
 
-You can also provide the user's details if you've collected them earlier in your service, to [prefill fields on the payment page](/optional_features/prefill_user_details).
+You can [prefill the fields on the payment page](/optional_features/prefill_user_details) with the user's details if you collected that information:
```
m-green: @SandorArpa This pull request should be good to go now. Could you just double-check the facts in this short bullet list are still correct?

source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -0,0 +1,47 @@
...
+target="blank">API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more information.
+
+## Data collected by your GOV.UK Pay admin account
+
+After the user completes their payment, your [GOV.UK Pay admin
+account](https://selfservice.payments.service.gov.uk/login) will collect the user's:
+
+- email address - even if you've disabled collecting your users' email addresses
+- cardholder name
+- billing address - if you've enabled collecting your users' billing addresses
```
katiebates: should this say "disabled" [billing address]? also @SandorArpa is this still true, since the edge case you found - i.e. front end overwrites address to null/ empty if collecting billing address is disabled in self-service.

SandorArpa: We store the billing address fields if any/all sent to the API. If the service is set NOT to collect billing address when (and only when) the user clicks continue on the card details page to confirm the payment, the billing address details are overwritten to empty fields in connector by frontend. If the charge stays in created state (eg the payment never been materialised) or it's being cancelled on the card details page then the billing address is not overwritten in connector.

source/payment_flow/index.html.md.erb
```diff
@@ -120,6 +120,11 @@ they have completed their payment attempt on GOV.UK Pay. This should not be
 JSON-encoded, because backslashes are invalid characters. With test accounts,
 you can [use HTTP for return URLs instead of HTTPS](/security/#https).
 
+You can [prefill the fields on the payment page](/optional_features/prefill_user_details) with the user's details if you collected that information:
+
+- earlier in your service, before you redirected your user to GOV.UK Pay
```
SandorArpa: I think it should say something along the lines 'before you created the charge'. When the user is redirected the charge was already created thus it's too late to include/exclude the billing address

m-green: As discussed, I've tweaked to just "earlier in the service" as this should be clear on its own.

source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -44,4 +44,6 @@ account](https://selfservice.payments.service.gov.uk/login) will collect the use
 
 - email address - even if you've disabled collecting your users' email addresses
 - cardholder name
-- billing address - if you've enabled collecting your users' billing addresses
+- billing address
```
m-green: @katiebates @SandorArpa I've tweaked lines 47 to 49 - does this look correct?

source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -0,0 +1,49 @@
...
+## Data collected by your GOV.UK Pay admin account
+
+After the user completes their payment, your [GOV.UK Pay admin
+account](https://selfservice.payments.service.gov.uk/login) will collect the user's:
+
+- email address - even if you've disabled collecting your users' email addresses
+- cardholder name
+- billing address
+
+If the user does not complete their payment, your GOV.UK Pay admin account will still collect the billing address. This will happen even if you've chosen not to collect your users' billing addresses.
```
SandorArpa: We don't collect it, we store it if it is sent to us. We don't display the billing address fields if the service decide that we shouldn't collect them. I'm not sure how this should be in the document...

m-green: Thanks Sandor - I've tweaked in https://github.com/alphagov/pay-tech-docs/pull/286/commits/932f05e0559a8b7ee8de1faf4eaab90b28436355.

PR #288, opened 2019-05-24
Payment links in Welsh

m-green:
### Context
Service teams can now create payment link payment pages in Welsh.

Update the 'Payment links' and 'Welsh language payment pages' pages with content about creating Welsh payment links.

### Guidance to review
Review and check if correct.

### DIFF
--- a/source/optional_features/welsh_language/index.html.md.erb
+++ b/source/optional_features/welsh_language/index.html.md.erb
@@ -5,12 +5,15 @@ weight: 132
 
 # Welsh-language payment pages
 
-GOV.UK Pay supports Welsh-language payment pages.
+GOV.UK Pay supports Welsh-language payment pages. Users will see Welsh text on the payment pages as they
+complete the [payment flow](/payment_flow).
 
-To use Welsh, you can include `"language": "cy"` in the body of a request to
-create a payment. Users will see Welsh text on the payment pages as they
-complete the [payment flow](/payment_flow).  If you do not specify a language,
-GOV.UK Pay will default to using English-language payment pages.
+To use Welsh, you can either:
+
+- include `"language": "cy"` in the body of a request to create a payment - if you're [integrating with the GOV.UK Pay API](/integrate_with_govuk_pay)
+- select **Create a payment link in Welsh** - if you're [creating a payment link](/payment_links) in the GOV.UK Pay admin tool
+
+If you do not specify a language when you create a payment, GOV.UK Pay will default to using English-language payment pages.
 
 Sign in to your [GOV.UK Pay
 account](https://selfservice.payments.service.gov.uk/login) and use the admin
@@ -19,10 +22,7 @@ tool to add a Welsh service name to your service.
 If you add a Welsh service name to your service, users will see this on Welsh-language
 payment pages instead of the English service name.
 
-When users make a payment, they will also [see a
-`description`](/payment_flow/#making-a-payment). If you use Welsh-language
-payment pages, you may also want to use Welsh text in the
-`description`.
+If you're integrating with the GOV.UK Pay API, you may also want to use Welsh text in the [`description`](/payment_flow/#making-a-payment). Users will see the Welsh `description` when they make a payment.
 
 GOV.UK Pay does not automatically translate emails into Welsh. Please [contact
 us](/support_contact_and_more_information/#contact-us) if you would like to send Welsh language emails to your users.
--- a/source/optional_features/welsh_language/index.html.md.erb
+++ b/source/optional_features/welsh_language/index.html.md.erb
@@ -5,15 +5,12 @@ weight: 132
 
 # Welsh-language payment pages
 
-GOV.UK Pay supports Welsh-language payment pages. Users will see Welsh text on the payment pages as they
-complete the [payment flow](/payment_flow).
+GOV.UK Pay supports Welsh-language payment pages.
 
-To use Welsh, you can either:
+You can also [create payment links in Welsh](/payment_links).
 
-- include `"language": "cy"` in the body of a request to create a payment - if you're [integrating with the GOV.UK Pay API](/integrate_with_govuk_pay)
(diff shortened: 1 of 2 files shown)


source/optional_features/welsh_language/index.html.md.erb
```diff
@@ -19,10 +22,7 @@ tool to add a Welsh service name to your service.
 If you add a Welsh service name to your service, users will see this on Welsh-language
 payment pages instead of the English service name.
```
alexbishop1: If you’re creating a Welsh payment link, you need to already have a Welsh service name set beforehand to avoid the URL containing the English service name (if you add a Welsh service name later, it won’t change the URLs of existing links). I believe we’ve had contact from services that were caught out by this, so it may be worth highlighting.

source/optional_features/welsh_language/index.html.md.erb
```diff
@@ -19,10 +22,7 @@ tool to add a Welsh service name to your service.
...
 
-When users make a payment, they will also [see a
-`description`](/payment_flow/#making-a-payment). If you use Welsh-language
-payment pages, you may also want to use Welsh text in the
-`description`.
+If you're integrating with the GOV.UK Pay API, you may also want to use Welsh text in the [`description`](/payment_flow/#making-a-payment). Users will see the Welsh `description` when they make a payment.
```
alexbishop1: This sort of applies to payment links too — the title the service sets for the payment link becomes the description and they should set this as Welsh. Not sure if it’s worth calling this out because the payment links setup flow makes it pretty clear you should be entering Welsh text.

source/optional_features/welsh_language/index.html.md.erb
katiebates: yep, definitely a good call

m-green: Good plan! I'll have a rejig.

source/optional_features/welsh_language/index.html.md.erb
m-green: > I wonder if it might be worth having separate sections for Welsh payments created by an API integration and Welsh payment links to make it clear that different considerations apply to each (e.g. the `description` advice only directly applies to API integrations while setting a Welsh service name also affects the URL with payment links).

@katiebates @alexbishop1 Thanks for the feedback!

I followed up with Tracy, and we realised I was complicating things by trying to use the Welsh page to detail how to set up payments links. That's not really a job for the tech docs. So in https://github.com/alphagov/pay-tech-docs/pull/288/commits/b6ef9b803bb04160057750ce3ffd1f822a8c9e6f I've pared things back - so we simply mention payment links are available in Welsh, then guide users to https://www.payments.service.gov.uk/payment-links/ as normal.

Tracy will highlight the need to set up a Welsh service name *before* you create a Welsh payment link, in content inside the admin tool itself or on the payment links product page.

source/optional_features/welsh_language/index.html.md.erb
m-green: See my comment below - Tracy's going to add this to the admin tool or features page, so the content is closer to where users need it.

source/optional_features/welsh_language/index.html.md.erb
```diff
@@ -7,9 +7,9 @@ weight: 132
 
 GOV.UK Pay supports Welsh-language payment pages.
 
-To use Welsh, you can include `"language": "cy"` in the body of a request to
-create a payment. Users will see Welsh text on the payment pages as they
-complete the [payment flow](/payment_flow).  If you do not specify a language,
+You can also [create payment links in Welsh](/payment_links).
```
alexbishop1: Not too sure this sentence fits here because the rest of the page is almost entirely about an API integration. Maybe this sentence should be right near the end, or possibly the next paragraph could have some exposition to make it really clear it’s only applicable to an API integration.

source/payment_links/index.html.md.erb
```diff
@@ -16,5 +16,6 @@ API if your service:
 
 * does not have a development team that can integrate your service with GOV.UK Pay
 
-You can read more about payment links on [the product
-page](https://www.payments.service.gov.uk/payment-links/).
+You can create payment links in English or Welsh.
```
alexbishop1: I wonder if it will be clear to the reader what it actually means to create a payment link in English or Welsh. Maybe something more along the lines of “Each payment link can be configured to display either English or Welsh instructions to users” (but without the word “configured” because that’s horrible).

source/optional_features/welsh_language/index.html.md.erb
m-green: Thanks @alexbishop1. As this more of a structure/style question, I'll make a note of this for the tech writer 2ier to consider.

source/payment_links/index.html.md.erb
m-green: @alexbishop1 Similar to the above - I'll note this for the 2ier so they can consider.

source/payment_links/index.html.md.erb
```diff
@@ -5,8 +5,9 @@ weight: 45
 
 # Payment links
 
-You may want to use payment links rather than integrate with the GOV.UK Pay
-API if your service:
+Payment links allow you to receive online payments from your users without integrating with the GOV.UK Pay API. When a user accesses your payment link, they’ll be able to make an online card payment.
```
m-green: @katiebates Could you review lines 8-10? Following 2i we've added a short description of Payment links (lifted directly from the product guidance) to reduce the risk Alex flagged of confusing payment links with payment pages.

source/payment_links/index.html.md.erb
m-green: Our 2ier felt this was ok as is.

source/optional_features/welsh_language/index.html.md.erb
m-green: Our 2ier felt this was ok but recommended adding a brief intro to the payment links page (in https://github.com/alphagov/pay-tech-docs/pull/288/commits/1306a363b728c4ac9e3ef8d98254577747a11a2a), to help distinguish the two features.

PR #291, opened 2019-06-04
Clarify delayed capture

m-green:
### Context
Update the delayed capture page to explain what the feature is and why you might use it.

Two sentences added to the delayed capture page.

### Guidance to review
Please review for factual accuracy.

### DIFF
--- a/source/optional_features/delayed_capture/index.html.md.erb
+++ b/source/optional_features/delayed_capture/index.html.md.erb
@@ -1,15 +1,15 @@
 ---
-title: Delayed capture
+title: Delayed capture of payments
 weight: 133
 ---
 
-# Delayed capture
+# Delayed capture of payments
 
-You can control how long it takes GOV.UK Pay to take ('capture') the money from a user's bank account after your Payment Service Provider (PSP) has authorised a payment.
+You can control how long it takes GOV.UK Pay to take ('capture') a payment from your user's bank account after your Payment Service Provider (PSP) has authorised that payment.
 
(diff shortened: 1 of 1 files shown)


source/optional_features/delayed_capture/index.html.md.erb
```diff
@@ -5,7 +5,9 @@ weight: 133
 
 # Delayed capture
 
-GOV.UK Pay supports the delayed capture of payments.
+You can control how long it takes for GOV.UK Pay to transfer ('capture') a payment from a user's bank account.
```
katiebates: i'm having second thoughts about this wording as it could imply that we can control the time it takes for a payment to be transferred from a user's account to the merchant's bank account (2 days - and we can't change that). you can control when a payment is removed from a user's account, after it has been authorised...?

m-green: How about "... to **take** ('capture') a payment from a user's bank account"?

m-green: I agree that 'transfer' sounds too much like we're talking about when it will arrive in the merchant's account.

katiebates: 'take' works :)

m-green: 👍

m-green:
Hi @katiebates - this should be ready to merge now, but could you take a look and check you're happy with the final version? (And complete your review if so!)

PR #298, opened 2019-06-22
Update Direct Debit API documentation

m-green:
### Context
Add documentation about the updates to the Direct Debit API.

- Update the Direct Debit page
- Add a line in Payment Flow to link users to the Direct Debit page for recurring payments

### Guidance to review
Review and check if correct.

### DIFF
--- a/source/direct_debit/index.html.md.erb
+++ b/source/direct_debit/index.html.md.erb
@@ -5,7 +5,7 @@ weight: 80
 
 # Direct Debit
 
-You can use Direct Debit if you want to take recurring payments.
+From August 2019, you can use Direct Debit if you want to take recurring payments.
 
 You can take payments whenever you need to. You can take the same amount or a different amount each time.
 
@@ -71,11 +71,6 @@ When the API responds, you need to:
 - direct the user to the `next_url` from the API response
 - store the `mandate_id` from the API response, so you can use that ID to collect payments against the mandate later
 
-Refer to the <a
-href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
-target="blank">GOV.UK Pay API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more
-information on the API response.
-
 ### 3. The user provides their bank account details
 
 When your service redirects the user to `next_url`, the user will see a **Set up a Direct Debit** page. The user enters their bank account details and selects **Continue**.
@@ -86,15 +81,17 @@ GOV.UK Pay checks with GoCardless that the user’s account number and sort code
 
 The screen presents the user’s account details. The user grants their consent
 by selecting __Confirm__.  GOV.UK Pay sends an email to the user to confirm
-that their mandate has been set up. If the mandate then fails, the
-GOV.UK Pay team will send an email to update the user that their mandate could
-not be set up.
+that their mandate is being set up.
 
 ### 5. GOV.UK Pay securely passes the user's details on to GoCardless
 
 GoCardless submits a request to BACS to set up a mandate, and
 sends a mandate ID to GOV.UK Pay. At this point, you can send a payment
 request.
+
+If the mandate then fails, the
+GOV.UK Pay team will send an email to update the user that their mandate could
+not be set up.
 
 ### 6. GOV.UK Pay returns the user to your service using the `return_url`
 
@@ -111,13 +108,7 @@ You should use these `status` and `details` parameters to display an appropriate
 
 The `status` will be `pending` until the user's bank confirms the mandate. The mandate confirmation process takes 2 working days.
 
-Refer to the <a
-href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
-target="blank">GOV.UK Pay API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more
-information on the API response.
-
 You can also check the status of a mandate by signing into your GoCardless account through the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
-
 
 ## Collecting payments against a mandate
 
@@ -167,11 +158,6 @@ If the payment exists, the API response will include a `state` object. This incl
 - [`status`](#payment-statuses)
 - [`details`](https://developer.gocardless.com/api-reference/#events-payment-causes), which explains the reason for the `status`
 - `finished`, which will be `true` if GOV.UK Pay has successfully collected the payment
-
-Refer to the <a
-href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
-target="blank">GOV.UK Pay API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more
-information on the API response
 
 You can also check the status of a payment by signing into your GoCardless account through the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login).
 
--- a/source/direct_debit/index.html.md.erb
+++ b/source/direct_debit/index.html.md.erb
@@ -16,19 +16,19 @@ You cannot automatically schedule payments. If you need to take regular payments
 To use the Direct Debit feature, you must [request a GOV.UK Pay
 Direct Debit test account](/support_contact_and_more_information/#contact-us) from our support team.
 
-You must also set up an account with our Direct Debit supplier, GoCardless.
+### Set up an account with our Direct Debit supplier
 
 1. Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
 
 1. Select your account in the __My services__ page and follow the on-screen instructions.
 
-You must [contact us](/support_contact_and_more_information/#contact-us) to sign your GoCardless contract before you switch to live.
+You must [contact us](/support_contact_and_more_information/#contact-us) before you switch to live, so we can send you your GoCardless contract to sign.
 
 ## Set up a mandate
 
-Before you can take payments, your user needs to set up a Direct Debit mandate ('instruction') by entering their bank account details on a GOV.UK Pay page.
+Before you can take payments, your user needs to set up a Direct Debit mandate (also known as an 'instruction') by entering their bank account details on a GOV.UK Pay page.
 
-The mandate is your user's agreement that you can take payments from their bank account.
+A mandate is your user's agreement that you can take payments from their bank account.
 
 These steps happen in order.
 
@@ -36,42 +36,40 @@ These steps happen in order.
 
 API call: `POST /v1/directdebit/mandates`
 
-Include the following parameters in your API call:
+You must include the following required parameters in your API call:
 
-- `description`
 - `reference`
 - `return_url`
 
-#### description
+You can include the optional `description` parameter in your API call.
 
-`description` is an optional human-readable description of the mandate.
+#### reference (required)
 
-The description must be no longer than 255 characters, and must not contain URLs.
+`reference` is a reference number you want to associate with this mandate.
 
-#### reference
-
-`reference` is the reference number you want to associate with this mandate.
-
-The reference number does not need to be unique, and must not:
+The reference number does not need to be unique, and can be any data type as long as it does not:
 
 - be longer than 255 characters
 - contain URLs
 
-The format is variable.
+#### return_url (required)
 
-#### return_url
-
-`return_url` is an HTTPS URL on your site. GOV.UK Pay will direct your user to this url when they've finished setting up the mandate.
+`return_url` is an HTTPS URL on your site. GOV.UK Pay will direct your user to this URL when GOV.UK Pay have finished setting up the mandate.
 
 The URL should not be JSON-encoded, because backslashes are invalid characters in the API.
 
+#### description (optional)
+
+`description` is a human-readable description of the mandate.
+
+The description must be no longer than 255 characters, and must not contain URLs.
 
 ### 2. Direct the user to GOV.UK Pay
 
 When the API responds, you need to:
 
 - direct the user to the `next_url` from the API response
-- store the `mandate_id` from the API response, so you can use it to collect payments against the mandate later
+- store the `mandate_id` from the API response, so you can use that ID to collect payments against the mandate later
 
 Refer to the <a
 href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
@@ -80,21 +78,21 @@ information on the API response.
 
 ### 3. The user provides their bank account details
 
-When your service redirects the user to `next_url`, they’ll see a **Set up a Direct Debit** page. The user enters their bank account details and selects **Continue**.
+When your service redirects the user to `next_url`, the user will see a **Set up a Direct Debit** page. The user enters their bank account details and selects **Continue**.
 
-GOV.UK Pay checks with GoCardless that the user's account details are valid.
+GOV.UK Pay checks with GoCardless that the user’s account number and sort code match a real bank account.
 
 ### 4. GOV.UK Pay presents the user with a confirmation screen
 
 The screen presents the user’s account details. The user grants their consent
 by selecting __Confirm__.  GOV.UK Pay sends an email to the user to confirm
-that their mandate has been set up. If the mandate subsequently fails, the
+that their mandate has been set up. If the mandate then fails, the
 GOV.UK Pay team will send an email to update the user that their mandate could
 not be set up.
 
-### 5. GOV.UK Pay securely passes the user's details on to the Direct Debit provider
+### 5. GOV.UK Pay securely passes the user's details on to GoCardless
 
-The Direct Debit provider submits a request to BACS to set up a mandate, and
+GoCardless submits a request to BACS to set up a mandate, and
 sends a mandate ID to GOV.UK Pay. At this point, you can send a payment
 request.
 
@@ -109,9 +107,9 @@ The API response includes a `state` object. This includes the following paramete
 - [`status`](#mandate-statuses)
 - [`details`](https://developer.gocardless.com/api-reference/#events-mandate-causes), which explains the reason for the `status`
 
-You should use the `status` and `details` to display an appropriate final page to your user.
+You should use these `status` and `details` parameters to display an appropriate final page to your user.
 
-The `status` will be `pending` until the bank confirms the mandate. The mandate confirmation process takes 2 working days.
+The `status` will be `pending` until the user's bank confirms the mandate. The mandate confirmation process takes 2 working days.
 
 Refer to the <a
 href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
@@ -121,7 +119,7 @@ information on the API response.
 You can also check the status of a mandate by signing into your GoCardless account through the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/).
 
 
-## Collect payments against a mandate
+## Collecting payments against a mandate
 
 API call: `POST /v1/directdebit/payments`
 
@@ -129,29 +127,36 @@ Required arguments:
 
 ```
 {
- "amount"
- "description" # optional
- "reference"
- "mandate_id"
+   "amount"
+   "reference"
+   "mandate_id"
+}
+```
+
+Optional argument:
+
+```
+{
+   "description"
 }
 ```
 
 Once a mandate has a `status` of `pending`, you can make an API call to collect a payment against the mandate.
 
-The payment will be taken from your user's bank account in:
+GoCardless will take the payment from your user's bank account in:
 
 - 3 working days if the mandate's `status` is `pending`
 - 2 working days if the mandate's `status` is `active`
 
-It then takes a further 1 working day before the payment reaches the service’s bank account.
+It then takes a further one working day before the payment reaches your service’s bank account.
 
 GOV.UK Pay will send an email to the user giving them 2 working days' advance
 notification that payment will be collected against their mandate. The email will contain the mandate reference number.
 
-You can collect whenever you need to. You may collect between £1 and £5000 each
-time. [Contact us](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) if you need to collect more than £5000.
+You can collect payments whenever you need to. You may collect between £1 and £5000 each
+time. [Contact us](https://docs.payments.service.gov.uk/support_contact_and_more_information/#contact-us) if you need to collect more than £5000 at any one time.
 
-### Get information about a payment
+### Getting information about a payment
 
 The call to check the status of a payment with the GOV.UK Pay API is:
 
@@ -160,7 +165,7 @@ The call to check the status of a payment with the GOV.UK Pay API is:
 If the payment exists, the API response will include a `state` object. This includes the following parameters:
 
 - [`status`](#payment-statuses)
-- [`details`](https://developer.gocardless.com/api-reference/#events-payment-causes), which is a reason for the status from GoCardless
+- [`details`](https://developer.gocardless.com/api-reference/#events-payment-causes), which explains the reason for the `status`
 - `finished`, which will be `true` if GOV.UK Pay has successfully collected the payment
 
 Refer to the <a
@@ -170,11 +175,11 @@ information on the API response
 
 You can also check the status of a payment by signing into your GoCardless account through the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login).
 
-## Generate a list of mandates or payments
+## Generating a list of mandates or payments
 
-You can search mandates or payments by using the API.
+You can use the API to search mandates or payments and return a list of results.
 
-### Search mandates
+### Searching mandates
 
 The call to search mandates with the GOV.UK Pay API is:
 
@@ -184,7 +189,7 @@ For example:
 
 `GET /v1/directdebit/mandates?from_date=2018-10-01T13:30:00Z&to_date=2018-10-14T15:00:00Z&reference=12345&status=pending`
 
-Some of the `queryparameters` you can use to search are:
+The query parameters you can use to search are:
 
 - `reference`
 - `status`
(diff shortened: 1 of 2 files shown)


source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
 </div>
 
-### Differences compared to card payment statuses
-
-The states `created`, `started` and `submitted` are all represented as
-a single state: `STARTED`.
-
-Direct Debit payments can have the non-terminal state `PENDING`.
-
 ## Reference numbers
```
m-green: @katiebates Can I double-check this section will still be correct as is? (It's not all shown here: [here's the link to the live section](https://docs.payments.service.gov.uk/direct_debit/#reference-numbers))

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+When the API responds, you need to:
 
-The response contains a `next_url`.
+- direct the user to the `next_url` from the API response
+- store the `mandate_id` from the API response, so you can use it to collect payments against the mandate later
 
-#### 2. Your service directs the user to the `next_url`
+Refer to the <a
+href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
+target="blank">GOV.UK Pay API browser</a> or the <a href="https://github.com/alphagov/pay-publicapi/blob/master/swagger/swagger.json" target="blank">Swagger file</a> for more
```
m-green: Links to the API browser need to be updated before publishing.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+The format is variable.
 
-#### 1. GOV.UK Pay creates the agreement
+#### return_url
 
-API call:  `POST /v1/agreements/`
+`return_url` is an HTTPS URL on your site. GOV.UK Pay will direct your user to this url when they've finished setting up the mandate.
 
-Your service must specify:
+The URL should not be JSON-encoded, because backslashes are invalid characters in the API.
```
alexbishop1: We should probably just fix the code to allow backslashes…

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+information on the API response.
 
-#### 3. The user provides their account details
+### 3. The user provides their bank account details
 
-For example, account number and sort code for Direct Debit.
+When your service redirects the user to `next_url`, they’ll see a **Set up a Direct Debit** page. The user enters their bank account details and selects **Continue**.
 
-GOV.UK Pay makes an API call to our Direct Debit supplier to verify account
+GOV.UK Pay makes an API call to GoCardless to verify the user's account
```
alexbishop1: When I firt read this, I thought it was referring to creating the mandate with GoCardless rather than just validating the bank account details. Maybe it could be something like: “GOV.UK Pay checks with GoCardless that the user entered details for a real bank account.”

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
 by selecting __Confirm__.  GOV.UK Pay sends an email to the user to confirm
 that their mandate has been set up. If the mandate subsequently fails, the
 GOV.UK Pay team will send an email to update the user that their mandate could
 not be set up.
 
-#### 5. GOV.UK Pay securely passes the user's details on to the Direct Debit provider
+### 5. GOV.UK Pay securely passes the user's details on to the Direct Debit provider
 
 The Direct Debit provider submits a request to BACS to set up a mandate, and
 sends a mandate ID to GOV.UK Pay. At this point, services can send a payment
```
alexbishop1: Should probably use “you” rather than “services” here.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
 request.
 
-#### 6. GOV.UK Pay returns the user to your service using the `return_url`
+### 6. GOV.UK Pay returns the user to your service using the `return_url`
+
+### 7. Call the API to check the status of the mandate
+
+API call: `GET /v1/directdebit/mandates/<mandateid>`
+
+This is to display an appropriate screen to the user.
```
alexbishop1: I don’t understand what this means.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+The API response includes a `state` object. This includes the following parameters:
 
-API call: `GET /v1/agreements/<AGREEMENT-ID>`
+- [`status`](#mandate-statuses)
+- [`details`](https://developer.gocardless.com/api-reference/#events-mandate-causes), which explains the reason for the `status`
 
-This is to display an appropriate screen to the user. For a Direct Debit
-payment, this will initially be in a ‘pending’ state until the bank confirms
-the mandate. You can confirm this state by signing into the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/). The mandate confirmation process takes 2 working days.
+The `status` will initially be in a `pending` state until the bank confirms the mandate. The mandate confirmation process takes 2 working days.
```
alexbishop1: Not sure about “initially” — initially it will be `created`: it’s only `pending` after the user has provided their details.

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
+ "mandate_id"
+}
+```
 
 Once a mandate has been set up and a mandate ID is available, you can make an
 API call to collect a payment against the mandate.
 
-You do not need to wait until the mandate is confirmed to send a collection
-request to GOV.UK Pay.
+You do not need to wait until the mandate is confirmed to send a collection request to GOV.UK Pay.
```
alexbishop1: But you do need to wait until the user has entered their details (`pending`), which is probably worth highlighting here.

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
 
-When collecting a payment against a mandate, you must [supply a
-Reference number](#reference-numbers) and description.
-
-If a mandate does not exist prior to a payment request, it takes 4 working
+If a mandate does not exist prior to a payment request, it takes 3 working
```
alexbishop1: Depending on how you look at it, the mandate does exist, it’s just not been set up with the bank yet.

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
-working days before the payment reaches the service’s bank account.
+If a mandate has already been set up, it takes 2 working days to collect a
+payment each time a subsequent one is requested. It then takes a further 1
+working day before the payment reaches the service’s bank account.
 
-GOV.UK Pay will send an email to the user giving them 3 working days' advance
+GOV.UK Pay will send an email to the user giving them 2 working days' advance
 notification that payment will be collected against their mandate. The email will contain the mandate reference number.
 
 You can collect at any time interval. You may collect between £1 and £5000 each
```
alexbishop1: I think “at any time interval” could just be “whenever you want” (or something).

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
+
+`GET /v1/directdebit/payments?<queryparameters>`
+
+For example:
+
+`GET /v1/directdebit/payments?from_date=2018-10-01T13:30:00Z&to_date=2018-10-14T15:00:00Z&reference=12345&status=success`
+
+Some of the `queryparameters` you can use to search are:
+
+- `reference`
```
alexbishop1: I think `reference` is a substring search for payments too.

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
+    <td><code>active</code></td>
+    <td>The PSP has confirmed that the mandate has been succesfully set up.</td>
+    </tr>    
+    <tr>
+    <td><code>inactive</code></td>
+    <td>You cannot collect payments from the mandate, for example because the user cancelled it.
+    </tr>    
+    <tr>
+    <td><code>cancelled</code></td>
+    <td>When the user entered their bank account information, they answered <b>No</b> to <b>Are you allowed to authorise Direct Debits on this account?</b>.</td>
```
alexbishop1: Maybe this could be vaguer and just say the user did not finish entering their details so the mandate was never set up?

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
 <div style = "overflow-x:auto;">
   <table style = "width:100%">
     <tr>
-    <th nowrap>Finished state</th>
+    <th nowrap>State</th>
     <th>Description</th>
+    </tr>
+    <td nowrap><code>pending</code></td>
+    <td>The payment has been requested, but the PSP is not yet able to confirm whether
+the payment is complete.</td>
```
alexbishop1: This can be as simple as the payment hasn’t been attempted yet. Maybe this could just be something like: “The payment has been requested but it is not yet known whether or not it will be successful.”

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
+    <td>The payment succeeded.</td>
     </tr>
-    <td><code>FAILED</code></td>
-    <td>Payment failed.</td>
+    <td><code>failed</code></td>
+    <td>The payment failed.</td>
     </tr>
-    <td><code>CANCELLED</code></td>
+    <tr>
+    <td><code>cancelled</code></td>
```
alexbishop1: I’m not actually sure if we have this functionality yet.

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
...
-
 ## Reference numbers
 
 You can control what appears on paying users’ bank account statements.  You'll
-enter this when you set up your account with the Direct Debit supplier. This
-applies to both one-off and variable Direct Debit payments.
+enter this when you set up your account with GoCardless.
 
-The  Direct Debit supplier will generate a mandate reference number, which is
+GoCardless will generate a mandate reference number, which is
```
alexbishop1: It might be worth mentioning here that it’s a Direct Debit scheme requirement that this reference be given to user and that it may appear on their bank statement.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
-    <td> Cancelled by the payer. </td>
-    </tr>
-    <tr>
-    <td nowrap><code> ACTIVE </code></td>
-    <td nowrap><code> CANCELLED </code></td>
-    <td> Cancelled by the payer. </td>
-    </tr>
-  </table>
-</div>
+1. Select **Link GoCardless Merchant Account** and follow the on-screen instructions.
```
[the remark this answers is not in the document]
stephencdaly: It can also be done with a test gocardless account if they want to test against a sandbox gocardless account. There are 3 different types of account they could have "test", "Gocardless test" and "Gocardless live".

The instructions are out of date though as this changed recently, now they select the account and then on the dashboard for that account they are presented with instructions for linking the account and a big green button to press.

Screen Shot 2019-06-25 at 10 30 33

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+### 1. GOV.UK Pay creates the mandate
 
-POST /v1/payments
+API call: `POST /v1/directdebit/mandates`
 
+Include the following parameters in your API call:
 
-You cannot take further payments from users using this method without creating
-a new payment. This requires users to re-enter their account details.
+- `description`
```
stephencdaly: description can be optional

source/direct_debit/index.html.md.erb
```diff
@@ -255,96 +133,197 @@ Required arguments:
  "amount"
  "description"
```
stephencdaly: description is now optional

stephencdaly: maybe it was good if the field specification was in the same format as we've given for the create mandate request.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
-scheduled Direct Debit payment, by sending an API call on a set schedule.
-Users would be emailed each time a payment is collected against their mandate.
+The reference number does not need to be unique, and must not:
 
-### Variable Direct Debit payment process
+- be longer than 255 characters
+- contain URLs
 
-These steps happen in order.
+The format is variable.
```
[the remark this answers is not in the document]
m-green: This is borrowed from https://docs.payments.service.gov.uk/making_payments/#reference - it's about teams being able to use string/number/etc as they like, rather than being restricted to one format. Let me know if it doesn't make sense here though.

source/direct_debit/index.html.md.erb
m-green: Thanks for the 'description' info!
I'd definitely like to make the field specifications consistent, but it may be too much for this iteration. I'll check at the 2i stage - but we'll make everything consistent in the next iteration in any case.

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
 
 # Direct Debit
 
-Before reading this section, you should be familiar with [making
-payments](/making_payments). You should also
-have [integrated your service](/integrate_with_govuk_pay/#how-to-integrate-with-the-gov-uk-pay-api) with the GOV.UK Pay API.
+You can use Direct Debit if you want to take recurring payments. You can take the same amount or a different amount each time.
 
-## Synchronous and asynchronous payment methods
+You cannot automatically schedule payments. If you need to take regular payments, you could create your own scheduler to send API calls on a set schedule.
```
katiebates: Could this call out something more positive about this - e.g. You have flexibility about when you can take payments. You cannot automatically...

That frequency flexibility is a selling point!

source/direct_debit/index.html.md.erb
```diff
@@ -5,248 +5,126 @@ weight: 80
...
+1. Select **Link GoCardless Merchant Account** and follow the on-screen instructions.
 
-## One-off payments using Direct Debit
+You must [contact us](/support_contact_and_more_information/#contact-us) to sign your GoCardless contract before you switch to live.
 
-To take a single payment using Direct Debit, you must set up a Direct Debit mandate.
+## Set up a mandate
 
-1. Create an API key linked to your Direct Debit service.
+You need to set up a Direct Debit 'mandate' before you can take payments. A mandate is your user's agreement that you can take payments from their bank account.
```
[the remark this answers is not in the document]
katiebates: I wonder if the use of "you" here is a bit confusing.. since it's users who set up the mandate...

source/direct_debit/index.html.md.erb
katiebates: at the risk of hair splitting, i wonder if 'validate' is more accurate since GoCardless don't verify that the account details belong to the person - it's just validating that they're real account details.

source/direct_debit/index.html.md.erb
katiebates: although i can see that further down we say 'you can collect at any interval'

source/direct_debit/index.html.md.erb
katiebates: at the risk of being repetitive, i wonder if it's worth emphasising that more granular information will be available with the status

source/direct_debit/index.html.md.erb
m-green: @alexbishop1 Just to check I've understood this - what status would the mandate be if it hadn't been set up with the bank (ie a 3-day turnaround)? And what status would it be if it *had* been set up (ie a 2-day turnaround).

source/direct_debit/index.html.md.erb
m-green: I'm not sure about adding that again here - as you say, it'd be very repetitive because we'd need to repeat it for every status code. It feels like service teams should be understand that `details` is available from the mentions earlier in the docs?

source/direct_debit/index.html.md.erb
m-green: I'll probably leave this out for now if that's ok - it's probably best to avoid getting too far into the service team's responsibilities around the Direct Debit scheme in the docs.

source/direct_debit/index.html.md.erb
m-green: Thanks Katie - added "You can take payments whenever you need to" here to clarify.

source/direct_debit/index.html.md.erb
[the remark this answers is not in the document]
m-green: I've reverted to the original version of the instructions as I think they're still correct, based on @stephencdaly's comment.

source/direct_debit/index.html.md.erb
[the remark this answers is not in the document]
m-green: Thanks both - I've removed the quotes and clarified who sets up the mandate in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: Added in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: Good points, thanks! I've clarified in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c, including using 'valid'.

source/direct_debit/index.html.md.erb
m-green: Done! Updated in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c

source/direct_debit/index.html.md.erb
m-green: This phrasing was borrowed from the 'Payment flow' page, but I agree that it isn't very clear. I've updated in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c with clearer language from elsewhere in 'Payment flow'.

source/direct_debit/index.html.md.erb
m-green: Thanks - I've updated this in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: I've clarified this in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: I checked with @katiebates and I've updated this in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c so it's hopefully clearer.

source/direct_debit/index.html.md.erb
m-green: Tweaked  in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c - thanks!

source/direct_debit/index.html.md.erb
m-green: Thank you - added in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: Good idea - I've updated this in https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c.

source/direct_debit/index.html.md.erb
m-green: Removed it :) https://github.com/alphagov/pay-tech-docs/pull/298/commits/562c05d6cc099a67190a4bfe298a6acc6291794c

source/direct_debit/index.html.md.erb
```diff
@@ -5,285 +5,253 @@ weight: 80
...
-## Get started
-
-To use the Direct Debit feature, you [must request a GOV.UK Pay
+To use the Direct Debit feature, you must [request a GOV.UK Pay
 Direct Debit test account](/support_contact_and_more_information/#contact-us) from our support team.
 
-[Contact us](/support_contact_and_more_information/#contact-us) if you want to add Direct Debit to an existing integration.
-
-### Set up an account with the Direct Debit supplier
+You must also set up an account with our Direct Debit supplier, GoCardless.
```
alexbishop1: I’m worried someone could see this and go off to GoCardless’s website and try to set up an account. They don’t need to do this. They can set up an account as part of the account linking flow. Can we phrase this so it informs them they’ll need to set this up as part of the set-up process but not in a way that it could be interpreted as an instruction to follow now?

source/direct_debit/index.html.md.erb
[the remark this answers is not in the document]
alexbishop1: I see. Maybe we could say something along the lines of, “You can use any format that’s suitable for your service as long as you follow the rules above.”

source/direct_debit/index.html.md.erb
```diff
@@ -5,285 +5,253 @@ weight: 80
...
 
-You can collect payments of variable amounts at any time by using a variable agreement.
-For example you may want to collect:
+### 3. The user provides their bank account details
 
-* money each time a service is provided
+When your service redirects the user to `next_url`, they’ll see a **Set up a Direct Debit** page. The user enters their bank account details and selects **Continue**.
 
-* payments where the amounts vary according to service usage
+GOV.UK Pay checks with GoCardless that the user's account details are valid.
```
alexbishop1: I share Katie’s concern that a reader may infer that the details are definitely valid and that the mandate set-up won’t fail. Can we do something weaker like, “GOV.UK Pay checks with GoCardless that the user’s account number and sort code match a real bank account”?

source/direct_debit/index.html.md.erb
m-green: I could remove 'GoCardless', so it wouldn't be possible to read this sentence as a direct instruction? That would be closer to the original heading (which was '## Set up an account with the Direct Debit supplier'.)

source/direct_debit/index.html.md.erb
m-green: Sounds good! I'll add this.

source/direct_debit/index.html.md.erb
katiebates: sorry i missed this. all correct apart from reference numbers being shown in the GOV.UK Pay admin tool - that's the long term vision, but won't be available to begin with. plus we don't send those reference numbers to GC so they won't be available in their admin tool. so just available in the API for now.

m-green: Ah good to know - thanks Katie. Will remove that bit.

source/direct_debit/index.html.md.erb
```diff
@@ -5,285 +5,259 @@ weight: 80
...
 
-* payments where the amounts vary according to service usage
+### 4. GOV.UK Pay presents the user with a confirmation screen
 
-* regular payments where you do not know in advance how much each payment will be
+The screen presents the user’s account details. The user grants their consent
+by selecting __Confirm__.  GOV.UK Pay sends an email to the user to confirm
+that their mandate has been set up. If the mandate then fails, the
+GOV.UK Pay team will send an email to update the user that their mandate could
+not be set up.
```
alexbishop1: “GOV.UK Pay sends an email to the user to confirm that their mandate has been set up” makes it sound a bit too much like this can’t fail. We should says something along the lines of “GOV.UK Pay sends an email to the user to inform them that their mandate is being set up”.

Chronologically, the “If the mandate then fails, the GOV.UK Pay team will send an email to update the user that their mandate could not be set up” bit happens after we’ve sent the details to GoCardless, so maybe that should be moved into the next step.

source/direct_debit/index.html.md.erb
```diff
@@ -5,285 +5,259 @@ weight: 80
...
 
-* a `return_url` where GOV.UK Pay will direct the user when they have finished setting up the agreement
+- [`status`](#mandate-statuses)
+- [`details`](https://developer.gocardless.com/api-reference/#events-mandate-causes), which explains the reason for the `status`
 
-* the `agreement_type` (`on demand`)
+You should use these `status` and `details` parameters to display an appropriate final page to your user.
 
-* a `reference` (optional) - your reference for this agreement
+The `status` will be `pending` until the user's bank confirms the mandate. The mandate confirmation process takes 2 working days.
```
alexbishop1: This might want to be qualified (it will only be `pending` if everything has gone well).

source/direct_debit/index.html.md.erb
```diff
@@ -5,285 +5,259 @@ weight: 80
...
-sends a mandate ID to GOV.UK Pay. At this point, services can send a payment
-request.
+```
+{
+   "description"
+}
+```
 
-#### 6. GOV.UK Pay returns the user to your service using the `return_url`
+Once a mandate has a `status` of `pending`, you can make an API call to collect a payment against the mandate.
```
alexbishop1: Do we want to explicitly state here that any attempt to collect a payment against a mandate can fail?

m-green: I'm not sure we need it here - the '## Collecting payments against a mandate' section talks about `status`, `detail` and `finished` states (and links to the various status/detail codes), so it should be clear that a payment could fail. Is that ok?

source/direct_debit/index.html.md.erb
m-green: Thanks Alex - will amend!

source/direct_debit/index.html.md.erb
m-green: I might have misunderstood the flow but - at this point in the chronology it seems the status could only be `pending`? Because the user has successfully filled in their details on GOV.UK Pay?

alexbishop1: The user might have cancelled or something. In theory I think it could also fail as soon as we send the details to GoCardless too.

At the point where the service is calling this API, all they know is the user has been returned from Pay. They won’t yet know if the user journey was successful.

source/direct_debit/index.html.md.erb
alexbishop1: Given we’re focusing on the happy path, I think that’s fine.

PR #300, opened 2019-06-26
Add character limits for parameters when prefilling fields

m-green:
### Context
We're missing the character limits for the API request parameters in https://docs.payments.service.gov.uk/optional_features/prefill_user_details/. 

Add the character limit for each parameter.

### Guidance to review
Please check if the changes are factually correct.

### DIFF
--- a/source/optional_features/prefill_user_details/index.html.md.erb
+++ b/source/optional_features/prefill_user_details/index.html.md.erb
@@ -29,9 +29,7 @@ For example:
 }
 ```
 
-All the parameters in `billing_address` are optional.
-
-The values must be no longer than:
+All the parameters in `billing_address` are optional, and these parameters' values must be no longer than:
 
 - 254 characters for `email`
 - 255 characters for `cardholder_name`
--- a/source/optional_features/prefill_user_details/index.html.md.erb
+++ b/source/optional_features/prefill_user_details/index.html.md.erb
@@ -31,14 +31,14 @@ For example:
 
 All the parameters in `billing_address` are optional.
 
-Each value must be no longer than:
+The values must be no longer than:
 
 - 254 characters for `email`
 - 255 characters for `cardholder_name`
-- 255 characters for `line1`, `line2` and `city` in `billing_address`
+- 255 characters each for `line1`, `line2` and `city` in `billing_address`
 - 25 characters for `postcode` in `billing_address`
 
-If you include the `country` parameter, it must be an <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
+If you include the `country` parameter, it must be an <a href="https://www.iso.org/obp/ui/#search/code/" target="blank">ISO 3166-1 Alpha-2 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
 
 Refer to the <a
 href="https://govukpay-api-browser.cloudapps.digital/#newpayment"
--- a/source/optional_features/prefill_user_details/index.html.md.erb
+++ b/source/optional_features/prefill_user_details/index.html.md.erb
@@ -38,7 +38,7 @@ The values must be no longer than:
 - 255 characters each for `line1`, `line2` and `city` in `billing_address`
 - 25 characters for `postcode` in `billing_address`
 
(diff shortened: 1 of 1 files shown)


source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -31,6 +31,13 @@ For example:
 
 All the parameters in `billing_address` are optional.
 
+Each value must be no longer than:
+
+- 254 characters for `email`
+- 255 characters for `cardholder_name`
+- 255 characters for `line1`, `line2` and `city` in `billing_address`
```
alexbishop1: I said something like this in a response to a support ticket and the user came back and asked if it meant 255 for each property or 255 characters in total across all those properties. Can we make it clear it’s per-field (which might be as simple as giving each of the four properties its own bullet point)?

source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -31,6 +31,13 @@ For example:
...
+- 25 characters for `postcode` in `billing_address`
+
 If you include the `country` parameter, it must be an <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
```
alexbishop1: I know you didn’t change this but since we’re here: it’s specifically ISO 3166-1 alpha-2 codes (we’ve had support queries about this).

source/optional_features/prefill_user_details/index.html.md.erb
m-green: Good point. How about:
`line1`, `line2` ~~and~~or `city`

I think giving each item its own bullet would unfortunately lead to too much duplication of '255'.

alexbishop1: I suppose “255 characters each for `line1`, `line2` and `city` in `billing_address`” would work.

m-green: Good plan! Added and tweaked slightly in https://github.com/alphagov/pay-tech-docs/pull/300/commits/d1ea6c6b7e6128ccb832c8fcf6cbf613ca5c09cf.

source/optional_features/prefill_user_details/index.html.md.erb
m-green: Good to know, thanks! Added in https://github.com/alphagov/pay-tech-docs/pull/300/commits/d1ea6c6b7e6128ccb832c8fcf6cbf613ca5c09cf.

source/optional_features/prefill_user_details/index.html.md.erb
```diff
@@ -31,7 +31,14 @@ For example:
...
 
-If you include the `country` parameter, it must be an <a href="https://www.iso.org/iso-3166-country-codes.html" target="blank">ISO 3166 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
+The values must be no longer than:
+
+- 254 characters for `email`
+- 255 characters for `cardholder_name`
+- 255 characters each for `line1`, `line2` and `city` in `billing_address`
+- 25 characters for `postcode` in `billing_address`
+
+If you include the `country` parameter, it must be an <a href="https://www.iso.org/obp/ui/#search/code/" target="blank">ISO 3166-1 Alpha-2 code</a>. If the parameter is invalid or missing, the __Country or territory__ field on the payment page will default to 'Great Britain'.
```
alexbishop1: I think you want `target="blank"` (open in a browsing context called `blank`) to be `target="_blank"` (open in a new browsing context). Also, it seems it’s always written “alpha-2” with a lower-case “a”. Finally, the default is actually the United Kingdom rather than Great Britain.

m-green: Ah thanks! We used `target="blank"` elsewhere because it worked - but looks like that will only work if you don't have an existing tab called 'blank'.
I thought it was 'Alpha-2' because ISO had written it that way on https://www.iso.org/obp/ui/#search/code/, but in retrospect that seems to be capitalisation just for their table heading? Will correct!

m-green: I've fixed those now in https://github.com/alphagov/pay-tech-docs/pull/300/commits/0ba6547dbd0d68f8fd659f736cd844a3118c1726.

alexbishop1: I hadn’t seen that table. I think you’re right they used capital letters at the beginning because they were treating it like the first word is a title. On https://www.iso.org/iso-3166-country-codes.html in running text they use “alpha-2” and “alpha-3” etc.

PR #302, opened 2019-07-10
Add content on capture_submit_time and capture_date 

m-green:
### Context
Teams who have Worldpay as their PSP can use `capture_submit_time` and `capture_date` from the API GET response to find out when Worldpay took a payment, to help them reconcile.

Update the 'Reporting' page to include content about `capture_submit_time` and `capture_date`.

### Guidance to review
Please check factual accuracy.

### DIFF
--- a/source/reporting/index.html.md.erb
+++ b/source/reporting/index.html.md.erb
@@ -89,6 +89,26 @@ have entered when making a payment:
     },
 ```
 
+### Checking when your PSP took a payment
+
+If your PSP is Worldpay, you can see when Worldpay took a payment from your user's bank account.
+
+The response contains:
+
+- `capture_submit_time` if we've asked Worldpay to take the payment from your user's bank account
+- `captured_date` if Worldpay have taken the payment
+
+For example:
+
+```
+  "settlement_summary": {
+    "capture_submit_time": "2016-01-21T17:15:000Z",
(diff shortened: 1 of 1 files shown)


source/reporting/index.html.md.erb
```diff
@@ -89,16 +89,36 @@ have entered when making a payment:
     },
 ```
 
+### Checking when your PSP took a payment
+
+If your PSP is Worldpay, you can see when Worldpay took a payment from your user's bank account.
```
alexbishop1: Card, not bank account

m-green: Hi Alex - we think it's best to stick with "bank account" here, as using "card" might leave some ambiguity about whether the user has actually had the payment taken from them. Hope that's ok.

alexbishop1: Hmm. For a credit card, the money won’t have been taken from the user’s bank account at that point (and might never be at all depending on how the user pays their credit card bill).

I could live with just “your user’s account”.

m-green: Ah good point - I've changed to "your user's account" as suggested in https://github.com/alphagov/pay-tech-docs/pull/302/commits/e1f09a4321a173975596da6ade8547c5cf51c8e3. Please approve if everything looks ok now!

PR #313, opened 2019-08-20
PP-5519 Update Docs With New P-Codes

RoryMalcolm:
- Updates the documentation with  new P Codes regarding the failure of
Direct Debit operations on PublicApi.


source/api_reference/index.html.md.erb
```diff
@@ -173,7 +173,8 @@ Common status codes are:
 | General | P0900 | Too many requests | Your service account is sending requests above the allowed rate. Please try the request again in a few seconds. |
 | General | P0920 | Request blocked by security rules | Our firewall blocked your request. See the [“Troubleshooting”](/troubleshooting) section for details. |
 | General | P0999 | GOV.UK Pay is unavailable | The GOV.UK Pay service is temporarily down. |
-| Direct Debit | P1103 | Invalid agreement state | You can only take payments from an agreement in `pending` or `active` states. |
+| Direct Debit | P1200 | Invalid mandate ID | Can't create payment. A mandate with the ID provided does not exist. |
```
stephencdaly: It looks like the text here should be something explaining why you're seeing the error, rather than the same text as the error. I'd ask @m-green what should go here.

RoryMalcolm: A payment cannot be taken as a mandate with the provided ID does not exist?

RoryMalcolm:
```suggestion
| Direct Debit | P1200 | Invalid mandate ID | A payment cannot be taken as a mandate with the provided ID does not exist. |
```

PR #320, opened 2019-09-04
Update `reference` parameter in documentation about searching for payments

m-green:
### Context
When you search for payments by `reference` using the API, the parameter now needs to be an exact match, rather than a partial string.

Update the Reporting and Direct Debit pages to update `reference`, and clarify as a whole which parameters are case sensitive or exact match only.

### Guidance to review

### DIFF
--- a/source/direct_debit/index.html.md.erb
+++ b/source/direct_debit/index.html.md.erb
@@ -176,13 +176,13 @@ For example:
 
 The query parameters you can use to search are:
 
-- `reference`
-- `status`
-- `bank_statement_reference`
+- `reference` - exact match only
+- `status` - exact match only
+- `bank_statement_reference` - exact match only
 - `email`
 - `name`
 
-All the parameters are case insensitive. You can search for partial values of `reference`, `email` and `name`.
+All the parameters are case insensitive.
 
 If you search for a specific mandate, all criteria you use must match. Otherwise your search will not return that mandate in the results.
 
@@ -204,7 +204,7 @@ The query parameters you can use to search are:
 - `status`
 - `mandate_id`
 
-All the parameters are case insensitive. You can search for partial values of `reference`.
+All the parameters are case insensitive but must be an exact match.
 
 If you search for a specific payment, all criteria you use must match. Otherwise your search will not return that payment in the results.
 
--- a/source/direct_debit/index.html.md.erb
+++ b/source/direct_debit/index.html.md.erb
@@ -176,13 +176,11 @@ For example:
 
 The query parameters you can use to search are:
 
-- `reference` - exact match only
-- `status` - exact match only
-- `bank_statement_reference` - exact match only
-- `email`
-- `name`
-
-All the parameters are case insensitive.
+- `reference` - case insensitive, exact match only
+- `bank_statement_reference` - case insensitive, exact match only
+- `email` - case insensitive
+- `name` - case insensitive
+- `state` - case sensitive, exact match only
 
 If you search for a specific mandate, all criteria you use must match. Otherwise your search will not return that mandate in the results.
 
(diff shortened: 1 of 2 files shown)


source/direct_debit/index.html.md.erb
```diff
@@ -176,13 +176,13 @@ For example:
 
 The query parameters you can use to search are:
 
-- `reference`
-- `status`
-- `bank_statement_reference`
+- `reference` - exact match only
+- `status` - exact match only
```
alexbishop1: I know it was already like this but it’s actually `state`, not `status`.

source/direct_debit/index.html.md.erb
```diff
@@ -176,13 +176,13 @@ For example:
...
+- `bank_statement_reference` - exact match only
 - `email`
 - `name`
 
-All the parameters are case insensitive. You can search for partial values of `reference`, `email` and `name`.
+All the parameters are case insensitive.
```
alexbishop1: I think `state` is case sensitive.

source/direct_debit/index.html.md.erb
```diff
@@ -204,7 +204,7 @@ The query parameters you can use to search are:
 - `status`
```
alexbishop1: Again, I know it was already like this but it’s actually `state`, not `status`.

source/direct_debit/index.html.md.erb
```diff
@@ -204,7 +204,7 @@ The query parameters you can use to search are:
...
 - `mandate_id`
 
-All the parameters are case insensitive. You can search for partial values of `reference`.
+All the parameters are case insensitive but must be an exact match.
```
alexbishop1: Once again, I think `state` is case sensitive.

source/direct_debit/index.html.md.erb
m-green: Thanks Alex - fixed in https://github.com/alphagov/pay-tech-docs/pull/320/commits/ec5793bfcb5ccd6af48b5f7977c3e98505ec0253.

source/direct_debit/index.html.md.erb
m-green: Thank you - fixed in https://github.com/alphagov/pay-tech-docs/pull/320/commits/ec5793bfcb5ccd6af48b5f7977c3e98505ec0253.

source/direct_debit/index.html.md.erb
m-green: Also fixed in https://github.com/alphagov/pay-tech-docs/pull/320/commits/ec5793bfcb5ccd6af48b5f7977c3e98505ec0253.

source/reporting/index.html.md.erb
```diff
@@ -149,22 +149,15 @@ An example search request:
...
-
-* `email`
-
-* `state`
-
-* `card_brand`
-
+* `reference` - case insensitive, exact match only
+* `email` - case insensitive
+* `card_brand` - case insensitive, exact match only
```
alexbishop1: Turns out the input for this doesn’t match the output in the response (e.g. you have to search for `master-card` to get `"card_brand": "Mastercard"`. But that disaster is probably not something to tackle now.

source/reporting/index.html.md.erb
```diff
@@ -149,22 +149,15 @@ An example search request:
...
+* `cardholder_name` - case insensitive
+* `state` - case sensitive, exact match only
 * `first_digits_card_number`
-
 * `last_digits_card_number`
```
alexbishop1: Might want to mention these are both exact matches too?

m-green: Good call - fixed in https://github.com/alphagov/pay-tech-docs/pull/320/commits/dfcde668ac365d847c513302e39b0f61b8c78056.

source/reporting/index.html.md.erb
m-green: Ah good to know - I've made a note to look at this later.

PR #330, opened 2019-09-17
Add `card_type` to documentation 

m-green:
### Context
We’re adding `card_type` to the information that teams can get about a payment through the API. It shows whether end users used a debit or credit card for the payment.

- Update https://docs.payments.service.gov.uk/reporting/ and http://localhost:4567/optional_features/digital_wallets/ with documentation about `card_type`.
- Remove the 'Card types' table from https://docs.payments.service.gov.uk/api_reference/, as it's no longer accurate.
- Fix a few trailing spaces.

### Guidance to review
Please check if factually accurate.


source/api_reference/index.html.md.erb
```diff
@@ -241,25 +241,6 @@ For example:
...
-| visa             | DEBIT  | Visa             |
-| visa             | CREDIT | Visa             |
-| master-card      | DEBIT  | Mastercard       |
-| master-card      | CREDIT | Mastercard       |
-| maestro          | DEBIT  | Maestro          |
-| american-express | CREDIT | American Express |
-| diners-club      | CREDIT | Diners Club      |
-| discover         | CREDIT | Discover         |
-| jcb              | CREDIT | JCB              |
-| unionpay         | CREDIT | Union Pay        |
```
alexbishop1: It turns out that when a service searches using the `card_brand` parameter, they have to use on the strings in the leftmost column (even though the `"card_brand"` property in the response will use one of the strings in the rightmost column!) so maybe it’s best not to remove this table altogether for now.

source/reporting/index.html.md.erb
```diff
@@ -87,9 +89,11 @@ have entered when making a payment:
     "city": "London",
     "country": "GB"
   }
-},
+}
 ```
 
+The `card_type` is `null` if we did not recognise which type of card your user made the payment with.
```
alexbishop1: Given the broad range of reasons why the `"card_type"` might be null, “we were unable to determine” feels more accurate to me that “we did not recognise” but I’m happy to defer to your judgement.

source/api_reference/index.html.md.erb
m-green: Thanks Alex - I'll add the table back in.

source/reporting/index.html.md.erb
m-green: If I remember right, I think we started with something along the lines of "we were unable to determine" - but I asked for some tech writer help to tweak it for style, and "we did not recognise" was the consensus. Let's give it a go and tweak it later if users aren't sure what we mean.

source/api_reference/index.html.md.erb
m-green: Added back in - https://github.com/alphagov/pay-tech-docs/pull/330/commits/afd6d75278d6986dd07fb6343712593f270b7eaa

PR #332, opened 2019-09-20
Improve documentation about resuming payments

m-green:
### Context
The 'Resume a payment' section of https://docs.payments.service.gov.uk/payment_flow/#resume-a-payment was potentially confusing as it referred to the actions as a 'feature'. It also no longer reflects the end user journey now we've made some improvements to it. 

Improve the 'Resume a payment' section, and remove the duplicate 'Incomplete payments' section.

### Guidance to review


source/payment_flow/index.html.md.erb
```diff
@@ -318,30 +318,23 @@ The `state` array within the JSON lets you know the outcome of the payment:
...
 more information about how you can integrate your service with GOV.UK Pay.
 
-## Resume a payment
+### Resuming an incomplete payment
 
-If your user starts a payment but does not complete it, they can resume that
-incomplete payment. An example of this could be if they use an internet
-browser’s ‘back’ button during a payment, and then go forward by selecting
-links on your website.
+You can send your user back to GOV.UK Pay if they arrive at your `return_url` before they've finished their payment. For example, your user selected their web browser’s **Back** button while making a payment, then selected **Forward**.
```
alexbishop1: I don’t think we should specifically mention the `return_url` as it’s unlikely the user will end up there; it’s more likely they’ll end up at an earlier page.

For example, imagine the flow of pages was like this:

https://www.example.service.gov.uk/apply/confirm-details (where the payment gets created) → GOV.UK Pay (various URLs) → https://www.example.service.gov.uk/apply/payment-complete (return URL)

Unless the user knows the return URL already, they’re unlikely to end up at https://www.example.service.gov.uk/apply/payment-complete without completing their payment. However, if they press the back button, they might end up at https://www.example.service.gov.uk/apply/confirm-details. At this point, the service could go, “Hang on, I’ve already created a payment for this user!” Then the service could check the status of the payment and offer to let the user resume if it’s unfinished (or continue with the rest of the journey if it’s finished).

Also, I think the change from “if they use an internet browser’s ‘back’ button during a payment, and then go forward by selecting links on your website” to “You can send your user back to GOV.UK Pay if they arrive at your `return_url` before they've finished their payment. For example, your user selected their web browser’s **Back** button while making a payment, then selected **Forward**” has lost some nuance. I don’t think it was trying to say that the user presses the back button and then the forward button in their browser. I think it was trying to say that if the user presses the back button in their browser that the service _could, by supporting the resume payment feature, show a link allowing the user to resume (go “forward” into) their payment.

source/payment_flow/index.html.md.erb
```diff
@@ -318,30 +318,23 @@ The `state` array within the JSON lets you know the outcome of the payment:
...
-If your service uses the resume payment feature, you will:
+You can only resume a payment if the `status` is one of the following:
 
-* minimise the number of expired payments
+- `created`
+- `started`
+- `submitted`
 
-* reduce the number of unnecessary new payments
+Resume the payment by [calling the GOV.UK API to find the payment](/reporting/#get-information-about-a-single-payment), then redirecting your user to the new `next_url` from the API response.
```
alexbishop1: Might be good to add that the service will want to check the status of the payment when they get it from the API. Maybe this sentence could be put before the list of resumable statuses (so it’s something along the lines of: “You can get the payment and check its status. If its status is one of these then there will be a `next_url` you can send the user to.”).

source/payment_flow/index.html.md.erb
```diff
@@ -318,30 +318,23 @@ The `state` array within the JSON lets you know the outcome of the payment:
...
-is a one-time URL. If a payment has already resumed using a `next_URL`, that
-URL will not be usable again.
+You should not reuse the `next_url` you received when you first created the payment. Each `next_url` has a unique token so you can only use it once.
 
-A payment cannot resume if it has a status of `success`, `failed`, `cancelled`
-or `error`. If your service tries to resume a payment of this type, your user
-will be sent to your service’s `return_url`. The `return_url` is the URL of a
-page on your service that GOV.UK Pay sends your user to after they have
-completed their payment attempt.
+We'll take your user to a screen that's appropriate for their payment’s current status. For example, if the `status` of the payment is `started`, we'll show the **Enter card details** page.
```
alexbishop1: The example might be more meaningful if the page the user gets sent to isn’t the first one in the flow. Maybe use `submitted` and the confirm page?

source/payment_flow/index.html.md.erb
m-green: Thanks Alex - makes sense. As discussed, I've scrunched this down to one sentence in https://github.com/alphagov/pay-tech-docs/pull/332/commits/7924b40b4e4d63cc7d3dd4078cc21be51d6e28b6 that I think does the job?

source/payment_flow/index.html.md.erb
m-green: Good plan. I've had a go at this in https://github.com/alphagov/pay-tech-docs/pull/332/commits/7924b40b4e4d63cc7d3dd4078cc21be51d6e28b6 but I'm still not sure I'm 100% there. Let me know what you think!

source/payment_flow/index.html.md.erb
m-green: 👍 Updated in https://github.com/alphagov/pay-tech-docs/pull/332/commits/7924b40b4e4d63cc7d3dd4078cc21be51d6e28b6

source/payment_flow/index.html.md.erb
```diff
@@ -320,19 +320,19 @@ more information about how you can integrate your service with GOV.UK Pay.
 
 ### Resuming an incomplete payment
 
-You can send your user back to GOV.UK Pay if they arrive at your `return_url` before they've finished their payment. For example, your user selected their web browser’s **Back** button while making a payment, then selected **Forward**.
+You can send your user back to GOV.UK Pay if they return to your service before they've finished their payment.
 
-You can only resume a payment if the `status` is one of the following:
+[Get information about the payment](/reporting/#get-information-about-a-single-payment) to check the payment's `status`.  the payment is incomplete if the `status` is one of the following:
```
alexbishop1: Need a capital letter at the beginning of that last sentence!

m-green: Oh dear - thanks for spotting!

PR #335, opened 2019-09-23
Improve our documentation about refunding payments

m-green:
### Context
We're improving the documentation for developers refunding payments using the GOV.UK Pay API, to make it more user-focused and concise.

General improvements to the content on https://docs.payments.service.gov.uk/refunding_payments/, including:
- restructuring and reshuffling the page into clearer sections 
- removing duplicated content
- updating 'Search refunds' to recent changes to https://docs.payments.service.gov.uk/reporting/#search-criteria 
- removing unneeded documentation about the admin tool
- fix links on pages that link to it

### Guidance to review


source/refunding_payments/index.html.md.erb
```diff
@@ -5,231 +5,106 @@ weight: 75
...
 
-GOV.UK Pay supports refunding payments. You can choose to refund part of a
-payment, or the full amount.
+You can refund all or part of a payment using the GOV.UK Pay API.
 
-Each payment has a refund status:
+You can also refund a payment by signing into the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login).
 
-<div style="height:1px;font-size:1px;">&nbsp;</div>
+Your payment service provider (PSP) will send refunds back to the source account of the original payment. You cannot refund a payment to another card or bank account.
```
mrlumbu: Referring to 'your PSP' might be confusing for services using Stripe?

source/refunding_payments/index.html.md.erb
```diff
@@ -5,231 +5,106 @@ weight: 75
...
-Go to the __Transactions__ section to see a list of transactions. For example,
-in the following image::
+`GET /v1/refunds/PAYMENT-ID`
 
-![](/images/transaction+list+image+4.png)
+The JSON response will include a `status` field.
 
-In this list, you can select an individual payment to see details of that
-payment, including any previous refunds.
+The `status` usually changes every 30 minutes. Do not check the status more than once every 5 minutes.
```
mrlumbu: Not sure about `The status usually changes every 30 minutes`?

source/refunding_payments/index.html.md.erb
```diff
@@ -239,70 +114,82 @@ If the payment has reached your bank account, your PSP will take the refund amou
...
-8601](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard).  For example, a valid input would be
-`2015-08-13T12:35:00Z`.
+Dates must be in [ISO
+8601](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard) format. For example `2019-09-20T13:30:00Z`.
 
-#### Pagination
+### Splitting results into pages
 
-You can use the following query parameters for pagination:
+You can use the following query parameters to split results into pages:
```
mrlumbu: since this is tech docs I would keep `pagination` as all devs will understand that, unless there is some guidance I am not aware of

source/refunding_payments/index.html.md.erb
mrlumbu: `Your PSP` maybe confusing for services integrating with Stripe?

m-green: This phrasing was already in the live docs (https://docs.payments.service.gov.uk/refunding_payments/#what-happens-next) - but happy to tweak it if it's not working? Maybe "**The** payment service provider" might work better?

source/refunding_payments/index.html.md.erb
m-green: Happy to change this. Is there a more accurate figure? Or we could say "It may take up to X for the `status` to change" if that's preferable?

mrlumbu: Sorry this was a bit unclear. I think the intent of the sentence is "Typically changes to refund status take 30 minutes to be reflected in GOV.UK Pay reporting" or something? This is actually quite PSP dependent, with some PSPs it will be more or less instantaneous. I guess the main purpose of this guidance is to give services a ball park figure for how regularly to check for updates, so maybe something along the lines of ☝️ might work better?

source/refunding_payments/index.html.md.erb
mrlumbu: Maybe we don't need to reference the PSP at all? 
`Refunds will be sent back to the source account of the original payment`?

source/refunding_payments/index.html.md.erb
m-green: Ah I see, thanks! So I think "It may take up to 30 minutes for `status` to change" is probably the right approach? I do take the point that it misses a bit of the nuance - in that it's the change being reflected, rather than the change itself. But from the user's perspective, it should get the facts across clearly - which is that they shouldn't worry if they don't see the change immediately.

source/refunding_payments/index.html.md.erb
mrlumbu: sounds good

source/refunding_payments/index.html.md.erb
mrlumbu: yep, that sounds good

PR #337, opened 2019-09-27
BAU: Add 4000000000000010 mock card number to testing docs

alexbishop1:
Add the 4000000000000010 mock card number to the testing docs, which card ID says is a Visa card for which it is unknown whether it is prepaid and for which it is unknown whether it is a credit or debit card but which is not a corporate card.


m-green:
Thanks for checking - this is is definitely ok for now, but I'd like to look at this table when we get a chance. I think the 'Card type' column could cause a bit of confusion now we have `card_type`, so we might need to tweak things a bit.

m-green:
p.s. could you add me as a reviewer to any docs updates (even minor ones), so I can check for style and run things through the 2i process if needed?

alexbishop1:
I shall ensure that even tiny doc changes are run past you first from now on, @m-green!

PR #345, opened 2019-10-24
Improve documentation about card brands

m-green:
### Context
The [card types table](https://docs.payments.service.gov.uk/api_reference/#card-types) is misleading and a little confusing (and oddly placed in the API reference page), especially now we've implemented `card_type` in payment information in addition to `card_brand`.

- Move the table out of the API reference page into the Reporting page, and include only needed information (including clarifying that the search value is different to the API response value for some card brands)
- Fix incorrect card types in the [mock card numbers table](https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers) 

### Guidance to review

### DIFF
--- a/source/reporting/index.html.md.erb
+++ b/source/reporting/index.html.md.erb
@@ -156,7 +156,6 @@ Some of the query parameters you can use to search for payments are:
 
 * `reference` - case insensitive, exact match only
 * `email` - case insensitive
-* `card_brand` - case insensitive, exact match only
 * `cardholder_name` - case insensitive
 * `state` - case sensitive, exact match only
 * `first_digits_card_number` - exact match only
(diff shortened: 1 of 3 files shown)


source/reporting/index.html.md.erb
```diff
@@ -166,6 +165,24 @@ If you search for a specific payment, all criteria you use must match. Otherwise
...
+
+- case insensitive and exact match only
+- different to the value in the API response for some card brands
+
+|`card_brand` to search for|`card_brand` in API response|
+|-|-|
+|`visa`|`Visa`|
+|`master-card`|`Mastercard`|
+|`maestro`|`Maestro`|
+|`american-express`|`AmericanExpress`|
```
alexbishop1: The API response is actually “American Express”

source/reporting/index.html.md.erb
```diff
@@ -166,6 +165,24 @@ If you search for a specific payment, all criteria you use must match. Otherwise
...
+|`diners-club`|`DinersClub`|
```
alexbishop1: The API response is actually “Diners Club”

source/reporting/index.html.md.erb
```diff
@@ -166,6 +165,24 @@ If you search for a specific payment, all criteria you use must match. Otherwise
...
+|`discover`|`Discover`|
+|`jcb`|`JCB`|
```
alexbishop1: The API response is actually “Jcb” (yeah, I know, right?)

source/reporting/index.html.md.erb
```diff
@@ -166,6 +165,24 @@ If you search for a specific payment, all criteria you use must match. Otherwise
...
+|`unionpay`|`UnionPay`|
```
alexbishop1: The API response is actually “Union Pay” (even though the real name of the card scheme doesn’t have a space!)

source/reporting/index.html.md.erb
m-green: Ah thank you @alexbishop1 - I did a blanket search-and-replace for spaces in the table, which is where all the spaces in the values went 😊

PR #363, opened 2019-12-18
Clarify delayed capture statuses and error codes

m-green:
### Context
The status codes in our [Delayed capture] docs are misleading or ambiguous. We also don't say what status a payment will have if the capture times out.

- Clarify the codes in the error codes table, and make them consistent with the descriptions in the [API reference](https://docs.payments.service.gov.uk/api_reference/#api-reference) page.
- Add the status for expired payments

### Guidance to review

### DIFF
--- a/source/optional_features/delayed_capture/index.html.md.erb
+++ b/source/optional_features/delayed_capture/index.html.md.erb
@@ -40,7 +40,7 @@ There are 4 possible responses:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-Your service must send the delayed capture request within 120 hours (5 days) of creating the payment, regardless of how long your user takes to complete the payment flow. Otherwise [the payment will expire](https://docs.payments.service.gov.uk/payment_flow/#resuming-an-incomplete-payment).
+Your service must send the delayed capture request within 120 hours (5 days) of creating the payment, regardless of how long your user takes to complete the payment flow. Otherwise [the payment will expire](/payment_flow/#payment-expires).
 
 If your service has enabled payment receipt emails, GOV.UK Pay will send a
 payment receipt email to your user once your service’s capture request is
--- a/source/optional_features/delayed_capture/index.html.md.erb
+++ b/source/optional_features/delayed_capture/index.html.md.erb
@@ -31,18 +31,16 @@ There are 4 possible responses:
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-| Response Code  | Meaning                    |
-| -------------- | -------------------------- |
-|204             | Capture request accepted   |
-|404             | `paymentId` does not exist |
-|409             | Capture request rejected   |
-|500             | Downstream error           |
+|Response code | Meaning |
+|--|--|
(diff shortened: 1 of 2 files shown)


source/optional_features/delayed_capture/index.html.md.erb
```diff
@@ -31,18 +31,16 @@ There are 4 possible responses:
...
+|404| No payment matched the `paymentId` you provided. |
+|409| The payment cannot be captured because its `status` is not `capturable`. |
+|500| Something is wrong with GOV.UK Pay. Please [contact us](/support_contact_and_more_information/#contact-us). |
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-Your service must send the delayed capture request within 120 hours (5 days) of the
-payment creation, regardless of how long your user takes to complete the
-payment flow. Otherwise, it will expire.
+Your service must send the delayed capture request within 120 hours (5 days) of creating the payment, regardless of how long your user takes to complete the payment flow. Otherwise the payment will expire. An expired payment has a `status` of `failed` and returns a [P0020 API error](/api_reference/#api-error-codes).
```
alexbishop1: I think this might be a bit confusing. The P0020 API error will be in the response if a service queries the payment (along with the `status` of `failed`). But putting here could be interpreted as saying that if a service tries to capture the payment, the response will include P0020.

m-green: Thanks Alex - in https://github.com/alphagov/pay-tech-docs/pull/363/commits/2087b8e7721e53e2ed73eb64ff7058fed89cebc4 I've linked to the Payment flow section about expired payments instead.

source/optional_features/delayed_capture/index.html.md.erb
```diff
@@ -31,18 +31,16 @@ There are 4 possible responses:
...
+|404| No payment matched the `paymentId` you provided. |
+|409| The payment cannot be captured because its `status` is not `capturable`. |
+|500| Something is wrong with GOV.UK Pay. Please [contact us](/support_contact_and_more_information/#contact-us). |
 
 <div style="height:1px;font-size:1px;">&nbsp;</div>
 
-Your service must send the delayed capture request within 120 hours (5 days) of the
-payment creation, regardless of how long your user takes to complete the
-payment flow. Otherwise, it will expire.
+Your service must send the delayed capture request within 120 hours (5 days) of creating the payment, regardless of how long your user takes to complete the payment flow. Otherwise [the payment will expire](https://docs.payments.service.gov.uk/payment_flow/#resuming-an-incomplete-payment).
```
alexbishop1: This link should be to https://docs.payments.service.gov.uk/payment_flow/#payment-expires.

I just checked that link and it mentions the standard 90 minute expiry, which is going to be confusing, so I think that will need some tweaks too.

m-green: Sorry about that - fixed (and other changes discussed) in https://github.com/alphagov/pay-tech-docs/pull/363/commits/c94e1afb26f89ef2e9ae0117a19e3bd877329027.

PR #369, opened 2020-01-08
Add documentation about blocking prepaid cards

m-green:
### Context
Document how teams can block users from using prepaid cards.

Add guidance to our documentation about going live and taking payments from corporate cards.

### Guidance to review
Please check if accurate.


source/switching_to_live/index.html.md.erb
```diff
@@ -51,7 +51,20 @@ Choose which card types to accept by signing into the [GOV.UK Pay admin tool](ht
 
 If you use Worldpay, SmartPay or ePDQ, you must also make sure the same card types are selected in your PSP account.
 
-### If you use Worldpay
+### Block prepaid cards
+
+You can block users from paying with prepaid cards.
+
+You must register with Visa to get permission first. To register, either:
```
alexbishop1: I suppose this only applies if they accept Visa cards. Technically, they could choose not to. So maybe this should start, “If you accept Visa cards…”. That would also make the relationship a bit clearer (right now it almost sounds like Visa are the get to decide whether you can block prepaid cards at all).

source/switching_to_live/index.html.md.erb
```diff
@@ -51,7 +51,20 @@ Choose which card types to accept by signing into the [GOV.UK Pay admin tool](ht
...
+
+- [contact us](/support_contact_and_more_information/#contact-us) if you use Stripe
+- contact your PSP
+
+After you've registered with Visa, [contact us](/support_contact_and_more_information/#contact-us) to block prepaid cards on your account.
+
+We usually cannot block Mastercard cards, because we cannot always tell if they are prepaid or not.
```
alexbishop1: Do you think we need to be more explicit about the implications of this? That is, that if you block prepaid cards, most prepaid Mastercard cards will still be accepted.

source/switching_to_live/index.html.md.erb
```diff
@@ -51,7 +51,20 @@ Choose which card types to accept by signing into the [GOV.UK Pay admin tool](ht
 
 If you use Worldpay, SmartPay or ePDQ, you must also make sure the same card types are selected in your PSP account.
 
-### If you use Worldpay
+### Block prepaid cards
+
+You can block users from paying with prepaid cards.
```
alexbishop1: We might want to include some exposition about why a service might want to do this (fraud). And maybe push them towards only blocking prepaid cards if they have a problem with them (rather than just because they can).

m-green:
Thanks for the feedback @alexbishop1! I've addressed your comments in 
https://github.com/alphagov/pay-tech-docs/pull/369/commits/dad4439ccbd504f2304b45864c10e9a02e36a0a0 (and also decided to move the blocking cards content into a dedicated page, as it was getting a bit too much for the go-live page).

Let me know what you think.

source/optional_features/block_prepaid_cards/index.html.md.erb
```diff
@@ -0,0 +1,17 @@
+---
+title: Block prepaid cards
+weight: 140
+---
+
+# Block prepaid cards
+
+You can block users from paying with prepaid cards. You should only do this if users are making fraudulent payments with prepaid cards.
+
+You cannot block most Mastercard prepaid cards, because we usually cannot tell if Mastercard cards are prepaid or not.
```
alexbishop1: I still think this needs a more explicit explanation of what happens if you choose to block prepaid cards but someone tries to pay with a prepaid Mastercard card. Right now, it almost reads like you can’t turn the feature on if you want to block prepaid Mastercard cards.

source/optional_features/block_prepaid_cards/index.html.md.erb
```diff
@@ -0,0 +1,17 @@
...
+
+If you want to block Visa prepaid cards, you must register with Visa to get permission first. To register, either:
+
+- [contact us](/support_contact_and_more_information/#contact-us) if you use Stripe
+- contact your Payment Service Provider
+
+[Contact us](/support_contact_and_more_information/#contact-us) to block prepaid cards on your account.
```
alexbishop1: Being right at the end, I worry this step could get lost and a service might think they only have to contact their PSP (or us in the case of Stripe).

source/optional_features/block_prepaid_cards/index.html.md.erb
m-green: @alexbishop1 I chatted to Miriam and we feel this is probably ok as is, but we're going to see if our 2ier reads it the same way.

source/optional_features/block_prepaid_cards/index.html.md.erb
m-green: I can't see another way of structuring this unfortunately - but as above I'll see what our 2ier thinks.

source/optional_features/block_prepaid_cards/index.html.md.erb
```diff
@@ -0,0 +1,15 @@
...
+# Block prepaid cards
+
+You can block users from paying with prepaid cards. You should only do this if users are making fraudulent payments with prepaid cards.
+
+You cannot block most Mastercard prepaid cards, because we usually cannot tell if Mastercard cards are prepaid or not.
+
+[Contact us](/support_contact_and_more_information/#contact-us) and we’ll:
+
+- tell you how to get permission from Visa first - if you want to block Visa cards
+- block prepaid cards on your account
```
m-green: @miriamalice Following 2i and our discussion, I've simplified the above section for now. We can add more detailed steps in another iteration when we've fleshed them out. Let me know if it looks ok!

PR #374, opened 2020-01-28
Add documentation about payout times

m-green:
### Context
We're missing some of the detail about how long it takes a team to receive a payment ("payout") in their bank account after a user completes their payment.

- Add more detail about how long it takes to receive a payment
- Move these docs to the 'Finance and accounting systems' of the 'How to integrate' page, which is where we're currently meeting the need of reconciling payments

### Guidance to review

### DIFF
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -78,14 +78,17 @@ Read more about [reporting](/reporting/#reporting) and [refunding payments](/ref
 
 ### When you'll receive payments
 
-You'll receive a payment in your bank account within 2 business days of your user completing their payment.
+You'll receive the money in your bank account within 2 business days of your user completing their payment.
 
-It may take 3 business days if:
+It will take 3 business days if your user completed their payment:
 
-- your user completed their payment after 11pm
-- you use Stripe and your user completed their payment after midnight UTC
+- after midnight UTC on a weekday
+- at the weekend
+- during a bank holiday
 
-It may take longer if the payment is below the minimum 'payout', which is the total amount you must receive from your users before your payment service provider (PSP) will transfer payments to you. If you're using GOV.UK's PSP, the minimum payout is £1.
+It may take longer if the payment is below the ‘minimum payout’. For example, if you use GOV.UK Pay’s payment service provider (PSP), the minimum payout is £1. If your user completes a payment of 20p, you will not receive the money until you’ve received 80p more to reach the minimum payout.
+
+If your PSP is Worldpay, SmartPay or ePDQ, ask your PSP if you need to know your exact minimum payout or payment times.
 
 ## The GOV.UK Pay API
 
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -80,13 +80,9 @@ Read more about [reporting](/reporting/#reporting) and [refunding payments](/ref
 
 You'll receive the money in your bank account within 2 business days of your user completing their payment.
 
-It will take 3 business days if your user completed their payment:
+It will take 3 business days if your user completed their payment at the weekend or on a bank holiday.
 
-- after midnight UTC on a weekday
-- at the weekend
-- during a bank holiday
-
-It may take longer if the payment is below the ‘minimum payout’. For example, if you use GOV.UK Pay’s payment service provider (PSP), the minimum payout is £1. If your user completes a payment of 20p, you will not receive the money until you’ve received 80p more to reach the minimum payout.
+Your payment service provider (PSP) will only transfer money to you when the total of your users' payments is above your PSP's ‘minimum payout’. For example, if you use GOV.UK Pay’s payment service provider (PSP), the minimum payout is £1.
 
 If your PSP is Worldpay, SmartPay or ePDQ, ask your PSP if you need to know your exact minimum payout or payment times.
 
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -78,9 +78,9 @@ Read more about [reporting](/reporting/#reporting) and [refunding payments](/ref
 
 ### When you'll receive payments
 
-You'll receive the money in your bank account within 2 business days of your user completing their payment.
+You'll receive the money in your bank account within 2 working days of your user completing their payment.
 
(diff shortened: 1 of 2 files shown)


source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -76,6 +76,17 @@ You could also connect a customer-relationship management (CRM) or case manageme
 
 Read more about [reporting](/reporting/#reporting) and [refunding payments](/refunding_payments/#refunding-payments).
 
+### When you'll receive payments
+
+You'll receive a payment in your bank account within 2 business days of your user completing their payment.
```
miriamalice: This is probably true but some orgs have different arrangements with their PSP. This might be 2 business days, or might be more. e.g. FCO get weekly payouts

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -76,6 +76,17 @@ You could also connect a customer-relationship management (CRM) or case manageme
...
+
+It may take 3 business days if:
+
+- your user completed their payment after 11pm
```
miriamalice: not sure if this is by PSP and something we can definitely say...?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -76,6 +76,17 @@ You could also connect a customer-relationship management (CRM) or case manageme
...
+- you use Stripe and your user completed their payment after midnight UTC
```
miriamalice: if it's after midnight it would still be 2 working days with Stripe - its just that you're into the next working day

miriamalice: also, we might want to say that payouts on a saturday or sunday are processed on a monday and paid out on wednesday...?

m-green: Is that covered by using 'business days'?

source/integrate_with_govuk_pay/index.html.md.erb
m-green: I did consider adding this but I wonder if it's necessary? Because you'll know if you've set up a different arrangement with your PSP.

source/integrate_with_govuk_pay/index.html.md.erb
m-green: This was already in the docs (it's in the second paragraph below the screenshot here: https://docs.payments.service.gov.uk/payment_flow/#payment-successful). I think Katie and I added it a while back and we haven't seen any issues since, so is it to leave on that basis?

m-green:
@miriamalice I've committed our version from our discussion earlier, and also made some tweaks to the 'minimum payout' description following a tech writer crit. Could you check you're ok with this, and then I'll send it for 2i?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -76,6 +76,20 @@ You could also connect a customer-relationship management (CRM) or case manageme
...
+
+You'll receive the money in your bank account within 2 business days of your user completing their payment.
+
+It will take 3 business days if your user completed their payment:
+
+- after midnight UTC on a weekday
+- at the weekend
+- during a bank holiday
+
+It may take longer if the payment is below the ‘minimum payout’. For example, if you use GOV.UK Pay’s payment service provider (PSP), the minimum payout is £1. If your user completes a payment of 20p, you will not receive the money until you’ve received 80p more to reach the minimum payout.
```
miriamalice: We should clarify that this is £1 net, ie after transaction fees. So a test payment of £1 would actually be 87p net and below the payout. We'd need to adjust the example accordingly (or if we think it gets too long winded, then cut the example).

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -76,6 +76,20 @@ You could also connect a customer-relationship management (CRM) or case manageme
 
 Read more about [reporting](/reporting/#reporting) and [refunding payments](/refunding_payments/#refunding-payments).
 
+### When you'll receive payments
+
+You'll receive the money in your bank account within 2 business days of your user completing their payment.
+
+It will take 3 business days if your user completed their payment:
+
+- after midnight UTC on a weekday
```
miriamalice: this would mean it's on the next working day and so is on the regular payout schedule. ie if it's after midnight on monday, it's really tuesday, so payout is on thursday. So we can cut this line I think and just leave the weekend and bank holiday info?

source/integrate_with_govuk_pay/index.html.md.erb
m-green: We've amended this to midnight UTC now.

source/integrate_with_govuk_pay/index.html.md.erb
m-green: We've added detail about midnight and weekends following our discussion.

source/integrate_with_govuk_pay/index.html.md.erb
m-green: Think we're ok on this one following our chat, so resolving the conversation.

source/integrate_with_govuk_pay/index.html.md.erb
m-green: Ah, good point. It's a bit tricky to add this because we don't mention exact transaction fees in the tech docs - so I think I'll need to go with cutting the example.

I might need to rejig the text a bit though. Is it true to say "Your PSP will only transfer money to you when the total of that day's payments is above their ‘minimum payout’."?

source/integrate_with_govuk_pay/index.html.md.erb
m-green: I think this line was was there for users who need to know that payments made at midnight BST (11pm UTC) will take 2 days rather than 3?

m-green: As discussed - removed in https://github.com/alphagov/pay-tech-docs/pull/374/commits/b5ed8c144e88f9b620bb9b28f0664a6be40cc253.

source/integrate_with_govuk_pay/index.html.md.erb
m-green: @miriamalice Updated in https://github.com/alphagov/pay-tech-docs/pull/374/commits/b5ed8c144e88f9b620bb9b28f0664a6be40cc253 following our chat.

m-green:
@miriamalice This has been 2i'd now - could you green-tick this one too so I can merge it in?

PR #376, opened 2020-01-30
Add documentation about checking account settings during go-live

m-green:
### Context
When users create a live account from a test account, certain settings will not be carried over.

Add a new step to the [go live docs](https://docs.payments.service.gov.uk/switching_to_live/) about switching on/off the same settings in your live account as in your test account.

### Guidance to review

### DIFF
--- a/source/switching_to_live/index.html.md.erb
+++ b/source/switching_to_live/index.html.md.erb
@@ -31,7 +31,7 @@ You'll need to tell us:
 - your organisation's name and address
 - if you’ll be using GOV.UK Pay’s PSP Stripe, or another PSP
 
-Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/) and select __Request a live account__.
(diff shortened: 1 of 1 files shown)


source/switching_to_live/index.html.md.erb
```diff
@@ -63,7 +63,17 @@ If you use Worldpay and you want to accept American Express cards, you must cont
 
 You must send your American Express merchant ID to Government Banking or Worldpay if you’re a non-Crown merchant.
 
-## 4. Check your live account works
+## 4. Check your settings match your test account
+
+You need to make the following changes to your live account if you made them to your test account:
```
miriamalice: Technically they might have made some changes to their test account that they actually don't want in the live account. That's probably clear and fine to leave if we think it is, but it might be good to clarify that we mean - if you made these changes in test and you want them in live, you'll need to make the changes in live as well.  ?

m-green: @miriamalice As discussed we'll leave as is for now on the basis that most teams will want their live accounts to match their test accounts. 

This has been 2i'd now - could you green-tick and then I'll merge it in :)

PR #381, opened 2020-02-06
Add MOTO payments documentation

m-green:
### Context
We can now set up a teams' GOV.UK Pay services so they can take MOTO payments (payments by over the phone or by post).

- Add documentation about MOTO payments, including error codes
- Alphabetise the 'Optional feature' page so it's easier to navigate

### Guidance to review

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -150,7 +150,7 @@ Common status codes are:
 |--------------|------------|---------|-------|
 | Create payment | P0101 | Missing mandatory attribute | The request you sent is missing a required attribute. |
 | Create payment | P0102 | Invalid attribute value | The value of an attribute you sent is invalid. |
-| Create payment | P0196 | MOTO payments not allowed | Contact us about [setting up your service for Mail Order Telephone Order (MOTO) payments](/optional_features/moto_payments/).|
+| Create payment | P0196 | MOTO payments not allowed | Contact us about [setting up your service for Mail Order / Telephone Order (MOTO) payments](/optional_features/moto_payments/).|
 | Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid.|
 | Create payment | P0198 | Downstream system error | Internal error with GOV.UK Pay. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
 | Create payment | P0199 | Account error | There is a problem with your service account. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -150,7 +150,7 @@ Common status codes are:
 |--------------|------------|---------|-------|
 | Create payment | P0101 | Missing mandatory attribute | The request you sent is missing a required attribute. |
 | Create payment | P0102 | Invalid attribute value | The value of an attribute you sent is invalid. |
-| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid. |
+| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid. Check the JSON in your request body, or [contact us about Mail Order Telephone Order (MOTO) payments](/optional_features/moto_payments/) if you've used `moto: true` in your request.|
 | Create payment | P0198 | Downstream system error | Internal error with GOV.UK Pay. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
 | Create payment | P0199 | Account error | There is a problem with your service account. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
 | Find payment by ID | P0200 | `paymentId` not found | No payment matched the `paymentId` you provided. |
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -150,7 +150,8 @@ Common status codes are:
 |--------------|------------|---------|-------|
 | Create payment | P0101 | Missing mandatory attribute | The request you sent is missing a required attribute. |
 | Create payment | P0102 | Invalid attribute value | The value of an attribute you sent is invalid. |
-| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid. Check the JSON in your request body, or [contact us about Mail Order Telephone Order (MOTO) payments](/optional_features/moto_payments/) if you've used `moto: true` in your request.|
+| Create payment | P0196 | MOTO payments not allowed | Contact us about [setting up your service for Mail Order Telephone Order (MOTO) payments](/optional_features/moto_payments/).|
+| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid.|
 | Create payment | P0198 | Downstream system error | Internal error with GOV.UK Pay. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
 | Create payment | P0199 | Account error | There is a problem with your service account. Please [contact us](/support_contact_and_more_information/#contact-us), quoting the error code. |
 | Find payment by ID | P0200 | `paymentId` not found | No payment matched the `paymentId` you provided. |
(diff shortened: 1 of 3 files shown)


source/optional_features/moto_payments/index.html.md.erb
```diff
@@ -0,0 +1,44 @@
...
+We can set up your service so you can fill in GOV.UK Pay payment pages with your users’ details.
+
+This means you can take payments:
+
+- over the phone
+- by post
+
+These payments are also known as Mail Order Telephone Order (MOTO) payments.
+
+After we set up your service, you'll no longer be able to create payments where users fill in their details themselves. [Create a separate service](https://selfservice.payments.service.gov.uk/my-services/create) if you need to do this.
```
katiebates: MOTO gateway accounts do allow 'normal' user-led online payments

m-green: Ah sorry, I might have misunderstood this.
Would you be able to take 'normal' payments on Worldpay, ePDQ and Stripe? Or just Stripe?

source/api_reference/index.html.md.erb
```diff
@@ -150,7 +150,7 @@ Common status codes are:
 |--------------|------------|---------|-------|
 | Create payment | P0101 | Missing mandatory attribute | The request you sent is missing a required attribute. |
 | Create payment | P0102 | Invalid attribute value | The value of an attribute you sent is invalid. |
-| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid. |
+| Create payment | P0197 | Unable to parse JSON | The JSON you sent in the request body is invalid. Check the JSON in your request body, or [contact us about Mail Order Telephone Order (MOTO) payments](/optional_features/moto_payments/) if you've used `moto: true` in your request.|
```
oswaldquek: I wouldn't make any changes to P0197. There should be a new row with P0196 instead: https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/model/PaymentError.java#L22

source/optional_features/index.html.md.erb
```diff
@@ -5,22 +5,15 @@ weight: 130
...
+You can use the following optional features with GOV.UK Pay:
+
+* [add corporate card surcharges](/optional_features/corporate_card_surcharges/#corporate-card-surcharges) to payments
+* [add custom branding](/optional_features/custom_branding/#custom-branding-of-payment-pages) to payment pages
+* [add custom metadata](/optional_features/custom_metadata) to payments
+* [block prepaid cards](/optional_features/block_prepaid_cards/#block-prepaid-cards)
+* [delay capturing payments](/optional_features/delayed_capture/#delayed-capture-of-payments)
+* [enable Apple Pay and Google Pay](/optional_features/digital_wallets/#digital-wallets)
+* [prefill fields](/optional_features/prefill_user_details/) on the **Enter card details** page
+* [take payments over the phone ('MOTO' payments)](/optional_features/moto_payments)
```
oswaldquek: I'd probably say `take Mail Order Telephone Order (MOTO) payments` instead of `take payments over the phone` as we probably will have another "take payments over the phone" feature that uses Twilio (or something similar). What do you think @katiebates ?

source/optional_features/moto_payments/index.html.md.erb
oswaldquek: Yeah I think there's some confusion here. If Stripe is being used, MOTO and normal payments can be made. However as we discussed yesterday, for Worldpay and ePDQ where services can take MOTO payments by virtue of using a MID, we decided that it was best to use a gateway account for one of MOTO payments or normal payments. The reason is that if we want to enable gateway accounts using Worldpay and ePDQ to take MOTO and normal payments, we'd have to do work to deal with two MIDs on a single gateway account.

source/api_reference/index.html.md.erb
m-green: Sorry Oz, I spotted my mistake and corrected it just before you commented! Fixed in https://github.com/alphagov/pay-tech-docs/pull/381/commits/166aa4228ca517d9e8b29a520910499eeede697c

source/optional_features/moto_payments/index.html.md.erb
katiebates: we did this so that services can have MOTO and non-MOTO payments within one account, so reporting can be in one place.

source/optional_features/moto_payments/index.html.md.erb
```diff
@@ -0,0 +1,44 @@
...
+## Creating MOTO payments
+
+To complete a MOTO payment, you must collect the following from your user:
+
+- card number
+- card expiry date
+- card security code
+- name from their card
+- email address - if you’ve [enabled collecting your users’ email addresses](/integrate_with_govuk_pay/#collecting-your-users-39-billing-addresses)
+
```
katiebates: might need to mention that we don't collect billing address for MOTO payments (even if it's enabled on the gateway account), because it's not used as part of transaction processing

source/optional_features/moto_payments/index.html.md.erb
```diff
@@ -0,0 +1,44 @@
...
+1. [Create a payment](/making_payments/#making-payments) with the GOV.UK Pay API, and include `moto: true` in the request body.
+
+2. Go to the `next_url` included in the body of the response to your API call.
+
+3. Fill in your user’s information on the **Enter card details** page.
+
+4. Confirm the payment on the **Confirm your payment** page.
```
katiebates: maybe we should also say that the 3DS/ 3DS2 challenge won't appear as part of the payment journey?

m-green:
@katiebates @oswaldquek I've rejigged things quite a bit in https://github.com/alphagov/pay-tech-docs/pull/381/commits/4d1aa9669282c9a0f0641d1f68647ac42d40fe28 following our chat - hope it's more accurate now. Let me know if anything needs changing.

source/optional_features/index.html.md.erb
oswaldquek: this conversation still needs resolution. i'm happy if you don't think this line needs to be changed but let me know.

source/optional_features/moto_payments/index.html.md.erb
```diff
@@ -0,0 +1,57 @@
...
+
+- card number
+- card expiry date
+- card security code
+- name from their card
+- email address - if you’ve [enabled collecting your users’ email addresses](/integrate_with_govuk_pay/#collecting-your-users-39-billing-addresses)
+
+Do the following to create and complete the MOTO payment.
+
+1. If your PSP is Worldpay or ePDQ, make sure you're using an API key from your MOTO account in [your services](https://selfservice.payments.service.gov.uk/my-services).
```
oswaldquek: I think it might be helpful to somehow highlight the difference if you're using Stripe.

source/optional_features/index.html.md.erb
m-green: I think in this instance it's best to lead with the plainer language version, because I suspect not everyone will be familiar with the term 'MOTO'.
Definitely agree that this language will be confusing if introduce a Twilio-type feature, and we should definitely revisit the description if that happens. But in terms of what's available now, this description feels ok?

source/optional_features/moto_payments/index.html.md.erb
m-green: Just to check - do you mean we should say something about Stripe users needing to use API keys from their 'normal' account?

source/optional_features/index.html.md.erb
oswaldquek: Ok, let's settle with the plainer language version. For the text "('MOTO' payments)" would it be a good idea to link the 'MOTO' to something like https://www.expertmarket.co.uk/merchant-accounts/moto-merchant-accounts to explain what MOTO is then?
Ok, I think this description feels ok for what we have now. We also have "('MOTO' payments)" which makes it clear.

m-green: Ah sorry, I haven't expanded the acronym. It should say:
- take payments over the phone (Mail Order Telephone Order ('MOTO') payments)
I'll fix that. (And decide what to do about the double brackets...)

We wouldn't be able to link to that commercial site I'm afraid. But in terms of explanation of MOTO and the benefits, I expect that'll definitely be something Tracy covers in the product page update. (Then these tech docs help devs with the task of actually doing the thing, at the point they already know what MOTO is and that they need to set it up.)

source/optional_features/moto_payments/index.html.md.erb
oswaldquek: Hmm, I was thinking more in general terms of making services aware that with Stripe, they won't need separate MOTO gateway accounts.
Actually I think this is alright for now. I'm not sure if services are clever enough to know that the implication with Stripe is that they only need one gateway account.

PR #390, opened 2020-02-24
Clarify account structure and user accounts

m-green:
### Context
We've had questions from service teams about how to set up their services in Pay.

Rejig the content to help avoid confusion about team members having access to different services.

### Guidance to review

### DIFF
--- a/source/account_structure/index.html.md.erb
+++ b/source/account_structure/index.html.md.erb
@@ -14,14 +14,7 @@ GOV.UK Pay test and live accounts are free to create and maintain but your payme
 
 If you do integrate multiple services, you can treat them separately or as a
 combined single service within GOV.UK Pay and your payment service provider (PSP).
-This gives 3 possible account configurations. These configurations allow you
-to alter:
-
-* which bank accounts payments go to
-
-* the description that appear on your users’ bank statements
-
-* who has access to each service and which permissions they have for each service
+This gives 3 possible account configurations.
 
 As an example, take 2 services called ‘parking permits’ and ‘parking fines’:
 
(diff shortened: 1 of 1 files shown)


m-green:
Hi @miriamalice - I've rethought this following our chat last week. I think there's still wider work we need to do on the page as a whole, but do you think this small change might help - where I've added (or technically moved) content about what need account type 1 meets? This means that every account type has a description of how it relates to user accounts.

source/account_structure/index.html.md.erb
```diff
@@ -35,6 +28,12 @@ both GOV.UK Pay and the PSP.
 In this setup, payments to services can go to separate bank accounts because
 each service has a PSP account.
 
+You can:
+
+- give access for each service to different staff members
+- report separately on each of your services
+- optionally use different names on the GOV.UK Pay payment pages
```
miriamalice: Should we clarify - 'different service names'?

source/account_structure/index.html.md.erb
```diff
@@ -35,6 +28,12 @@ both GOV.UK Pay and the PSP.
 In this setup, payments to services can go to separate bank accounts because
```
miriamalice: Would it be helpful to add this into the list of 'you can' things below?

source/account_structure/index.html.md.erb
m-green: Good plan - added!

source/account_structure/index.html.md.erb
m-green: Also a good plan - have added (and rejigged the bullet order so it better matches reality).

PR #393, opened 2020-03-03
Add setting maximum transaction amount in Worldpay settings

m-green:
### Context
Some payments are causing errors because they're higher than the 'Maximum transaction limit' in the team's Worldpay settings.

Add setting 'Maximum transaction amount' during the go-live process for Worldpay.

### Guidance to review

### DIFF
--- a/source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
+++ b/source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
@@ -49,11 +49,15 @@ If __Test Mode__ is at the bottom of the left-hand menu, switch to __Production 
 
 10. Enter your XML password into the __Password__ field in your GOV.UK Pay account.
 
-11. In the __Payment service__ section of your Worldpay account, set __Capture delay (days)__ to __Off__ (not __0__).
+11. Go to the __Payment service__ section of your Worldpay account.
+
+12. Set __Capture delay (days)__ to __Off__ (not __0__).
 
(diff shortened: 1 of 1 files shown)


source/switching_to_live/set_up_a_live_worldpay_account/index.html.md.erb
```diff
@@ -49,11 +49,15 @@ If __Test Mode__ is at the bottom of the left-hand menu, switch to __Production
...
 
-11. In the __Payment service__ section of your Worldpay account, set __Capture delay (days)__ to __Off__ (not __0__).
+11. Go to the __Payment service__ section of your Worldpay account.
+
+12. Set __Capture delay (days)__ to __Off__ (not __0__).
 
     If you do not do this, you may take money from your users before they've confirmed their payment.
 
-12. In your GOV.UK Pay account, select __Save credentials__.
+13. Set __Maximum Transaction Amount__ to the highest payment you will take. The maximum amount is 10,000,000 pence (£100,000).
```
stephencdaly: I think that they need to request that Worldpay do this for them, I don't think it's editable. Do we need to specify that? Or can they usually edit it themselves?

m-green: Sorry, you're right! I misread the thread. I've fixed in https://github.com/alphagov/pay-tech-docs/pull/393/commits/8cc2ec95c56255e3948d6e6d0235549b49b2babd.

PR #395, opened 2020-03-03
Add content about end user bank statement descriptions

m-green:
### Context
We need to be clearer about how teams can change the description that appears on end users' bank statements.

- Add content on the 'Payment flow' page
- Tweak the content on the 'Account structure' page

### Guidance to review

### DIFF
--- a/source/account_structure/index.html.md.erb
+++ b/source/account_structure/index.html.md.erb
@@ -92,4 +92,8 @@ Within each gateway account for the PSP, you can edit:
 
 * which of your bank accounts your revenue goes to
 
-You should contact your PSP to find out more.
+You can either:
+
+* [contact us](/support_contact_and_more_information/#support) - if your PSP is Stripe
+
+* contact your PSP
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -208,6 +208,11 @@ will then:
 
 Find out [when you'll receive the payment](/integrate_with_govuk_pay/#when-you-39-ll-receive-payments).
 
+You can change what appears on your user's bank statement by:
+
+- [contacting us](/support_contact_and_more_information/#support) - if your PSP is Stripe
+- contacting your PSP
+
 ### Confirmation email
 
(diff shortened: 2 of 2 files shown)


source/account_structure/index.html.md.erb
```diff
@@ -92,4 +92,8 @@ Within each gateway account for the PSP, you can edit:
 
 * which of your bank accounts your revenue goes to
 
-You should contact your PSP to find out more.
+You can either:
+
+* [contact us](/support_contact_and_more_information/#support) - if your PSP is Stripe
```
PippaClarkGDS: Looks good to me - one thing is maybe it should be: * [contact us](/support_contact_and_more_information/#support) (if your PSP is Stripe) ... so having "if your PSP is Stripe" in brackets. A dash looks like there's more longer information to be read rather than just an aside bit of information, if that makes sense

source/payment_flow/index.html.md.erb
```diff
@@ -208,6 +208,11 @@ will then:
 
 Find out [when you'll receive the payment](/integrate_with_govuk_pay/#when-you-39-ll-receive-payments).
 
+You can change what appears on your user's bank statement by:
+
+- [contacting us](/support_contact_and_more_information/#support) - if your PSP is Stripe
```
PippaClarkGDS: Looks good to me - one thing is maybe it should be: - [contacting us](/support_contact_and_more_information/#support) (if your PSP is Stripe) ... so having "if your PSP is Stripe" in brackets. A dash looks like there's more longer information to be read rather than just an aside bit of information, if that makes sense

source/account_structure/index.html.md.erb
m-green: Thanks Pippa - I've amended now.

source/payment_flow/index.html.md.erb
m-green: Amended in https://github.com/alphagov/pay-tech-docs/pull/395/commits/0352ef03a2e7dcc0561d31071b8b41f324b32062.

PR #398, opened 2020-03-10
Add docs about when to search payments

m-green:
### Context
If teams search payments at (say) midnight at the end of each day, and use an end time of midnight, they may miss some payments - because the API may not return fully up-to-date information about payment events from the last hour.

Add docs to our section about [integrating finance and accounting systems](https://docs.payments.service.gov.uk/integrate_with_govuk_pay/#finance-and-accounting-systems) on using an end date at least one hour before your search time, so you make sure to get up-to-date information about payment events up to that time.

### Guidance to review

### DIFF
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -67,7 +67,7 @@ using the API.
 
 For example, you could automatically fetch data about the outcome of payment journeys. You could import that into your finance system so payments can be reconciled against bank transaction information.
 
-If you [generate a list of payments](/reporting/#generate-a-list-of-payments-search-payments) at the end of each day, you should use a `to_date` of at least one hour before you generate the list. This is because the API may not return up-to-date information about payment events from the last hour.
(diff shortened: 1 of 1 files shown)


source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -67,6 +67,8 @@ using the API.
 
 For example, you could automatically fetch data about the outcome of payment journeys. You could import that into your finance system so payments can be reconciled against bank transaction information.
 
+If you [generate a list of payments](/reporting/#generate-a-list-of-payments-search-payments) at the end of each day, you should use a `to_date` of at least one hour before you generate the list. This is because the API may not return up-to-date information about payment events from the last hour.
```
miriamalice: This is correct, but I think, based on conversations with LAs, that they think about it the other way round. ie they know what time they want the to_date to be because it's set for other systems that rely on this data (ie midnight, or 5PM) and they need to run the report an hour later (ie at 1AM or 6PM). Rather than saying, you want to run a report at midnight so set the time for 11PM. Does that make sense...?

m-green: Good point, thanks @miriamalice! I've clarified in the latest commits (and tweaked a bit for clarity).

PR #400, opened 2020-03-12
Add docs about search being asynchronous

m-green:
### Context
Teams should not use the [search payments](https://docs.payments.service.gov.uk/reporting/#generate-a-list-of-payments-search-payments) API call during payment flow to check the status of a payment, because search is asynchronous so the payment status returned may not be up-to-date. 

Add docs to:
- Payment flow
- Reporting

We do not have documentation about the 'Get events for a payment' API call. To keep this an agile change, I've linked to the API browser for now, and I'll add documentation about that call to the tech docs later.
 
### Guidance to review

### DIFF
--- a/source/reporting/index.html.md.erb
+++ b/source/reporting/index.html.md.erb
@@ -145,7 +145,7 @@ You can also sign in to the [GOV.UK Pay admin tool](https://selfservice.payments
 
 See the <a href="https://govukpay-api-browser.cloudapps.digital/#searchpayments" target="blank">GOV.UK Pay API browser</a> for more details
 
-You must not search payments using the API to check a payment's status during [payment flow](/payment_flow). Search results are asynchronous so the payment status may not be up-to-date.
+Do not search payments using the API to check a payment's status during [payment flow](/payment_flow). Search results are asynchronous so the payment status may not be up-to-date.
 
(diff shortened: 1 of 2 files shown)


source/reporting/index.html.md.erb
```diff
@@ -145,6 +145,8 @@ You can also sign in to the [GOV.UK Pay admin tool](https://selfservice.payments
 
 See the <a href="https://govukpay-api-browser.cloudapps.digital/#searchpayments" target="blank">GOV.UK Pay API browser</a> for more details
 
+You must not search payments to check a payment's status during [payment flow](/payment_flow). Search results are asynchronous so the payment status may not be up-to-date.
```
heathd, line 148: I would not say "search payments" but rather "use the search payments endpoint".. don't know if this is consistent with other parts of the documentation though.

```suggestion
You must not use the search payments endpoint to check a payment's status during [payment flow](/payment_flow). Search results are asynchronous so the payment status may not be up-to-date.
```

m-green: Ah good point. The style would be "search payments using the API" - that will also make it consistent with https://github.com/alphagov/pay-tech-docs/pull/398/files so thanks for spotting!

PR #409, opened 2020-04-15
Add Direct Debit test account details

m-green:
### Context
Service teams can test their Direct Debit integration in their test account by using a mock sort code and account number.

Add a 'Test a mock bank account' section to our [Direct Debit documentation](https://docs.payments.service.gov.uk/direct_debit/#direct-debit).

### Guidance to review

### DIFF
--- a/source/direct_debit/index.html.md.erb
+++ b/source/direct_debit/index.html.md.erb
@@ -243,9 +243,9 @@ This would return payments 501 to 1000.
 
 ## Test a mock bank account
 
-To test confirming bank account details as a user, use the following with your test account:
+To test entering bank account details as a user, use the following with your test account:
 
-- `200000` as the sort code
+- `20 00 00` as the sort code
 - `55779911` as the account number
 
 ## Mandate statuses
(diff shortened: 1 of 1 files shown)


source/direct_debit/index.html.md.erb
```diff
@@ -241,6 +241,13 @@ You could then get a single page of these mandates by running:
 
 This would return payments 501 to 1000.
 
+## Test a mock bank account
+
+To test confirming bank account details as a user, use the following with your test account:
```
alexbishop1, line 246: Not sure “confirming” is the right word here.

source/direct_debit/index.html.md.erb
```diff
@@ -241,6 +241,13 @@ You could then get a single page of these mandates by running:
...
+
+- `200000` as the sort code
```
alexbishop1, line 248: Maybe write the sort code as “20 00 00”. The style guide at https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#bank-details says to put spaces in sort codes and we accept that format on the payment pages.

m-green: Ah good point! I'll fix that, thanks Alex.

source/direct_debit/index.html.md.erb
m-green: Apologies, something went wrong here... this should be "entering". I'll fix now.

PR #414, opened 2020-04-23
Clarify apple pay doesn't need Worldpay setting

heathd:
### Context

GOV.UK Pay fully decrypts the apple pay payload before passing to worldpay, so
there's no need to enable any special settings in worldpay.

We had a question about this from a partner service who wanted confirmation as
to whether this setting was needed. It's mentioned in Worldpay's docs:

https://beta.developer.worldpay.com/docs/wpg/mobilewallets/applepay#setup

Explicitly state that worldpay config changes not needed

### Guidance to review

Possibly we don't need to explain why here. I would be ok with that :-). I used the same language to describe this setting as is used in Worldpay's docs.

### DIFF
--- a/source/optional_features/digital_wallets/index.html.md.erb
+++ b/source/optional_features/digital_wallets/index.html.md.erb
@@ -15,6 +15,8 @@ Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.
 
 In your __Settings__, select the __Digital wallet__ tab in __Card types__. In the next screen, enable Apple Pay.
 
+You do not need to ask Worldpay to enable Apple Pay on your Worldpay account.
+
 ## Enable Google Pay
 
 To enable Google Pay, you must enter a unique Google Pay merchant ID into your GOV.UK Pay account.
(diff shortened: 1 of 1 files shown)


source/optional_features/digital_wallets/index.html.md.erb
```diff
@@ -15,6 +15,10 @@ Sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.
 
 In your __Settings__, select the __Digital wallet__ tab in __Card types__. In the next screen, enable Apple Pay.
 
+You do not need to ask Worldpay to enable Apple Pay on your Worldpay account
+to use this feature, because GOV.UK Pay fully handles the Apple Pay
```
m-green, line 19:
```suggestion
, because GOV.UK Pay fully handles the Apple Pay
```

Just for conciseness, as I think the meaning is still clear without this.

m-green: I know this is a familiar theme from me, but... I think we can lose everything from "because". That's not because the reason isn't important - it's because the reason is fully implied through just the instruction. We can say just "You do not need to ask Worldpay to enable Apple Pay in your Worldpay account", and the clear implication is "because it will work without you doing that".

heathd:
ok I've updated in line with your comments Mark and squashed the commits

PR #420, opened 2020-05-14
Clarify Direct Debit availability

m-green:
### Context and changes
Minor changes to the Direct Debit page to clarify that Direct Debit functionality is not currently available to central government departments in England or the NHS.

### Guidance to review


m-green:
@miriamalice I've reversed the order of https://github.com/alphagov/pay-product-page/pull/230/files because that version could be read as 'the NHS in England or government departments in England'. Hope that makes sense - if so, worth reversing on the product page too.

PR #428, opened 2020-05-21
Update review dates and fix factual issues (part 1)

m-green:
### Context
We're doing a fact check of all documentation pages so we can add Daniel The Manual Spaniel review dates to the documentation as a whole.

Makes factual fixes and adds review dates to the first set of documentation pages we've checked.

### Guidance to review
Please check it all looks ok.

### DIFF
--- a/source/index.html.md.erb
+++ b/source/index.html.md.erb
@@ -2,7 +2,7 @@
 title: Documentation – GOV.UK Pay
 weight: 10
 last_reviewed_on: 2020-03-09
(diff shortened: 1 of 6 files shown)


m-green:
First set of fact-check-a-thon changes!

- @miriamalice Please check the HTTPS change in Making Payments looks ok (the words are lifted directly from the integration page)
- @stephencdaly Please check the broken links you spotted now look ok.

PR #433, opened 2020-06-02
Call out the updates and notifications

stevenjmesser:
It'd be useful to call out the automated notifications of incidents, so that we can point to it when engaging with service teams.

### DIFF
--- a/source/support_contact_and_more_information/index.html.md.erb
+++ b/source/support_contact_and_more_information/index.html.md.erb
@@ -5,7 +5,7 @@ weight: 160
 
 # Support
 
-You can get notifications on GOV.UK Pay’s status to your email, phone, RSS reader or website, by subscribing on our [status page](https://payments.statuspage.io/#).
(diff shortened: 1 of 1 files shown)


source/support_contact_and_more_information/index.html.md.erb
```diff
@@ -5,8 +5,7 @@ weight: 160
 
 # Support
 
-You can subscribe to our public status page for updates on [the status of our
-system](https://payments.statuspage.io/).
+You can receive notifications and updates of incidents by email, text message, RSS or webhook from [our public status page](https://payments.statuspage.io/).
```
m-green, line 8:
```suggestion
You can get notifications on GOV.UK Pay’s status to your email, phone, RSS reader or website, by subscribing on our [status page](https://payments.statuspage.io/#).
```

Thanks for this @stevenjmesser! I've tweaked slightly for style, and switched to the means of delivery (as I wasn't sure if everyone would know what a webhook is). Let me know what you think!

stevenjmesser: That's a much better version of what I was trying to do!

PR #434, opened 2020-06-02
Add link to Stripe payout reporting

m-green:
### Context
We're adding payout reporting for Stripe to the GOV.UK Pay admin tool.

Add a link in the [finance and accounting section of the integrating docs](https://docs.payments.service.gov.uk/integrate_with_govuk_pay/#finance-and-accounting-systems).

### Guidance to review
Do not merge until the payout reporting is live.

### DIFF
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -69,6 +69,8 @@ You'll receive the money in your bank account within 2 working days of your user
 
 It will take 3 working days if your user completed their payment at the weekend or on a bank holiday.
 
+If your PSP is Stripe, you can see payments Stripe has made to your bank account by signing in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login).
+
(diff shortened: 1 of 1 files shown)


source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -69,6 +69,8 @@ You'll receive the money in your bank account within 2 working days of your user
 
 It will take 3 working days if your user completed their payment at the weekend or on a bank holiday.
 
+If your PSP is Stripe, you can see payments Stripe has made to your bank account by signing in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login).
```
miriamalice, line 72: we might need to say - payments and all the transactions that make up the payments? the breakdown of all transactions is useful for some people. or is the expectation that they access it and see what's in there? 

related - i don't think we've found a good way to show this to someone who is in a test account...

m-green: On the first point - yep, "payments" should be enough for users to click in. It's along the same lines of not over-explaining a service before they press 'Start'.

On the second point - do you think it's worth adding "You can only see payments for live accounts."?

PR #441, opened 2020-06-15
(Draft) PP-6616 Add payouts API documentation

m-green:
### Context
We're testing documentation about getting information about Stripe payouts through the API. 

Add new parameter for payout date to getting information about a payment or refund, and searching for payments or refunds.

### DIFF
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -80,11 +80,11 @@ Example response:
 ```json
 "settlement_summary": {
   ...
-  "psp_settled_date": "2016-01-22",
+  "settled_date": "2016-01-22",
 }
 ```
 
-`psp_settled_date` is either:
+`settled_date` is either:
 
 - the date that Stripe sent the payment to your bank account as part of a group of payments (‘payout’)
 - missing if Stripe has not yet sent the payment to your bank account
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -65,18 +65,31 @@ Read more about [reporting](/reporting/#reporting) and [refunding payments](/ref
 
 ### When you'll receive payments
 
-You'll receive the money in your bank account within 2 working days of your user completing their payment.
+You'll receive the money in your bank account within 2 working days of your user completing their payment. It will take 3 working days if your user completed their payment at the weekend or on a bank holiday.
 
-It will take 3 working days if your user completed their payment at the weekend or on a bank holiday.
-
-If your PSP is Stripe, you can sign in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login) to see:
-
-- payments ('payouts') that Stripe has made to your bank account
-- which of your users' payments are in each payout
-
-Your payment service provider (PSP) will only transfer money to you when the total of your users' payments is above your PSP's ‘minimum payout’. For example, if you use GOV.UK Pay’s PSP, the minimum payout is £1.
+Your payment service provider (PSP) will only transfer money to you when the total of your users' payments is above your PSP's ‘minimum payout’. For example if your PSP is Stripe, the minimum payout is £1.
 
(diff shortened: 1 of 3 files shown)


source/refunding_payments/index.html.md.erb
```diff
@@ -119,6 +119,22 @@ If the payment has reached your bank account, your PSP will take the refund amou
 - the next payment or payments you receive from any of your users, if your PSP is Stripe or ePDQ
 - your bank account, if your PSP is SmartPay or Worldpay
 
+If your PSP is Stripe, you can see when Stripe took the refund by making an API call to [get information about a payment](/reporting/#reporting-via-the-gov-uk-pay-api).
```
sfount, line 122: Quick comment while I'm looking through the rest more carefully, should this link to advice on `get information about a refund` or is it intentionally payment oriented? 

We do seem to have sections about getting refunds https://docs.payments.service.gov.uk/refunding_payments/#checking-the-status-of-a-refund, I notice it doesn't mention getting a single refund (although this is possible through the API (https://govukpay-api-browser.cloudapps.digital/#get-a-payment-refund)

m-green: Ah sorry - this is incorrect because I copied-and-pasted too quickly from another bit of the new docs.

I've updated now - could you check this is correct?

source/reporting/index.html.md.erb
```diff
@@ -233,6 +233,17 @@ These take inputs based on a subset of [the ISO
 8601 standard](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard).
 For example, a valid input would be `2015-08-13T12:35:00Z`.
 
+#### Filter by when your PSP sent payments to your bank account
+
+If your PSP is Stripe, you can use the following query parameters to filter by when your PSP sent payments to your bank account:
```
heathd, line 238: need to clarify how nulls a treated

if the payout hasn't happened yet is the payment included or excluded from the result set?

if I wanted to list all payments which haven't been settled, could I do that?

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -65,18 +65,31 @@ Read more about [reporting](/reporting/#reporting) and [refunding payments](/ref
...
-Your payment service provider (PSP) will only transfer money to you when the total of your users' payments is above your PSP's ‘minimum payout’. For example, if you use GOV.UK Pay’s PSP, the minimum payout is £1.
+If your PSP is Stripe, you can check when Stripe sent a payment to your bank account by making an API call to [get information about a payment](/reporting/#reporting-via-the-gov-uk-pay-api).
 
-If your PSP is Worldpay, SmartPay or ePDQ, ask your PSP if you need to know your exact minimum payout or payment times.
+Example response:
+
+```json
+"settlement_summary": {
+  ...
+  "psp_settled_date": "2016-01-22",
```
heathd, line 83:
```suggestion
  "settled_date": "2016-01-22",
```

suggest that we avoid the `psp_` prefix on fields in `settlement_summary`. The process of settlement is by definition something which is carried out by a PSP so it's redundant.

We already expose the following fields within `settlement_summary`:

```json
"settlement_summary": {
  "capture_submit_time": "2016-01-21T17:15:000Z",
  "captured_date": "2016-01-21"
}
```

these are described on the reporting page:

https://docs.payments.service.gov.uk/reporting/#checking-when-your-payment-service-provider-psp-took-a-payment

I think we should have one place where we describe the `settlement_summary`. Would it make more sense to consolidate the description of that functionality and cross reference it from here?

source/reporting/index.html.md.erb
sfount: That's a great point, filtering on created date doesn't run into this complexity as transactions are guaranteed to be created.

I'm not sure how it should be communicated but the answer to the functionality would likely be: 

> If the payout hasn't happened yet is the payment included or excluded from the result set?
Excluded, this wouldn't meet the search criteria

> If I wanted to list all payments which haven't been settled, could I do that?
No

jonathanglassman:
Closed as covered by https://github.com/alphagov/pay-tech-docs/pull/460

PR #442, opened 2020-06-15
Clarify MOTO payments

m-green:
### Context
We need to clarify that if teams want to use MOTO payments and they're using Worldpay and ePDQ as their PSP, they need to create a separate GOV.UK Pay service and get a separate merchant code.

Add the clarifications above.

### Guidance to review

### DIFF
--- a/source/optional_features/moto_payments/index.html.md.erb
+++ b/source/optional_features/moto_payments/index.html.md.erb
@@ -16,10 +16,9 @@ This means you can take payments:
 
 These payments are also known as Mail Order / Telephone Order (MOTO) payments.
 
-You cannot take MOTO payments:
+You cannot take MOTO payments if your payment service provider (PSP) is SmartPay.
 
-- if your payment service provider (PSP) is SmartPay
-- through a [payment link](https://www.payments.service.gov.uk/govuk-payment-pages/) - [contact us](/support_contact_and_more_information/#contact-us) if you'd like us to develop this feature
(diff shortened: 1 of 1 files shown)


source/optional_features/moto_payments/index.html.md.erb
```diff
@@ -16,14 +16,28 @@ This means you can take payments:
...
+## If your PSP is Worldpay or ePDQ
+
+If your PSP is Worldpay or ePDQ, you must first:
+
+- [create a new GOV.UK Pay service](https://selfservice.payments.service.gov.uk/create-service/register) for MOTO payments
+- contact your PSP and ask for a new merchant code for MOTO payments
+
+[Contact Government Banking](https://www.gov.uk/government/groups/government-banking-service-gbs) instead if you use them.
+
+You cannot use an existing GOV.UK Pay service or an existing merchant code.
```
miriamalice, line 33: existing e-commerce merchant code? they might have an existing MOTO code but that isn't in use in Pay. 

also they could have an existing MOTO service in Pay. -I guess the issue is not that it is existing,but that it separate from the online payments service.

miriamalice: should we explicitly say, if your PSP is Stripe, you can have MOTO and online payments on the same service? we sort of say it by indirectly by not including them in the above, but wonder if it might need clarification.

m-green: I think it's probably ok without it, but I'm in two minds... For testing purposes, I think it might be best to leave it as it is. We know that account structure can sometimes cause confusion, so I'd like to see if indirectly saying something's possible with accounts in this case (by not saying it isn't) leads to any support questions. What do you think?

PR #445, opened 2020-06-29
Add review dates and make factual fixes

m-green:
### Context
We're doing a fact check of all documentation pages so we can add Daniel The Manual Spaniel review dates to the documentation as a whole.

Makes factual fixes and adds review dates to the third set of documentation pages we've checked.

### Guidance to review

### DIFF
--- a/source/refunding_payments/index.html.md.erb
+++ b/source/refunding_payments/index.html.md.erb
@@ -1,5 +1,7 @@
 ---
 title: Refunding payments
+last_reviewed_on: 2020-06-29
+review_in: 6 months
 weight: 75
 ---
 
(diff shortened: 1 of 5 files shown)


source/refunding_payments/index.html.md.erb
```diff
@@ -95,15 +97,15 @@ If GOV.UK Pay could not successfully create the refund, the JSON response will c
 
 ## Checking the status of a refund
 
-`GET /v1/refunds/PAYMENT-ID`
+`GET /v1/payments/PAYMENT-ID/refunds/REFUND-ID`
```
alexbishop1, line 100: How was this wrong for so long? 😬

m-green: Good question! Did the endpoint change at some point maybe?

PR #446, opened 2020-07-07
Amend Pay tech docs navigation

jonathanglassman:
### Context

The current Pay tech docs navigation has a few characteristics that are not line with content best practice. Specifically:

- an "Optional features" section that does not tell users skimming the left hand nav what the section contains
- some of the titles are not active and verb-based
- the API reference section could be closer to the bottom of the navigation as it is supporting information
- the different types of payment (e.g. direct debit, moto) could have their own section 

- move content from "Optional features" section into new section "Customise your payment pages"
- create new section "Take different kinds of payment" with digital wallet, moto, and direct debit payments
- make titles more active and verb-based

### Guidance to review

- check that new navigation makes sense and content is in a logical position
- check that amended titles are ok

### Temp documentation site for review

https://paynavigation.london.cloudapps.digital/#gov-uk-pay-technical-documentation

### Jira ticket

https://payments-platform.atlassian.net/browse/PP-6721


jonathanglassman:
https://payments-platform.atlassian.net/browse/PP-6721

source/account_structure/index.html.md.erb
```diff
@@ -1,13 +1,13 @@
...
 review_in: 6 months
-weight: 95
+weight: 40
 ---
 
-# Account structure
+# How accounts work
 
-You can read more about [reporting](/reporting/#reporting).
+You can read more about [reporting](/reporting/#report-on-a-payment).
```
miriamalice, line 10: this feels a bit odd here - maybe it's a hangover from a previous structure?

miriamalice: Also the subheadings here 'editing Pay services' and 'editing PSP serivces' are a bt confusing. Maybe it's something about managing settings in Pay or managing settings on your PSP.

source/set_up_3dsecure/index.html.md.erb
```diff
@@ -2,7 +2,7 @@
 title: Set up 3D Secure
 last_reviewed_on: 2020-06-10
 review_in: 3 months
-weight: 110
+weight: 170
 ---
 
 # Set up 3D Secure (3DS)
```
miriamalice, line 8: it's also 3dS2 now as well (well, for Worldpay at least). Should we say 3DS and 3DS2? Or, we could say 'Strong Customer Authentication' which is the catchall for both but is a less familiar term.

source/take_different_payments/index.html.md.erb
```diff
@@ -0,0 +1,12 @@
+---
+title: Take different kinds of payment
+weight: 110
+---
+
+# Take different kinds of payment
```
miriamalice, line 6: I understand the logic of this but it feels a bit like these big features are hidden away a bit, particularly when smaller card features like blocking prepaid cards/delayed payments are in the main nav. Phone payments probably warrants it's own page (and we'll be adding more phone payment functionality at some point). Digital wallets could be a separate thing or nested somewhere.

source/versioning/index.html.md.erb
```diff
@@ -1,9 +1,9 @@
 ---
-title: Versioning
-weight: 130
+title: Stay up to date
+weight: 190
 ---
 
-# Versioning
+# Stay up to date
```
miriamalice, line 6: This is a much clearer title. if we're making it a more general thing about staying up to date should we include other things, e.g. subscribe to the statusio page? Or should we make it about staying up to date with the API?

source/account_structure/index.html.md.erb
jonathanglassman: agree, will change

source/set_up_3dsecure/index.html.md.erb
jonathanglassman: adding to backlog

source/take_different_payments/index.html.md.erb
jonathanglassman: @miriamalice that's a fair proposal. Currently the navigation has "Take a payment" as a top level heading in the left hand nav. To distinguish from these other types of payment, perhaps should amend this to something like "Take a card payment"?  Or does content in this section also apply to the other types of payment?

source/versioning/index.html.md.erb
jonathanglassman: I think specifying that it is keep the API up to date is the way forward for now. We can iterate further as ever.

source/account_structure/index.html.md.erb
```diff
@@ -1,11 +1,11 @@
...
-title: Account structure
+title: How accounts work
 last_reviewed_on: 2020-06-29
 review_in: 6 months
-weight: 95
+weight: 40
 ---
 
-# Account structure
+# How accounts work
```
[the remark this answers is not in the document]
jonathanglassman: @miriamalice have added to the tech docs backlog in Jira

jonathanglassman:
thanks @sfount for the github help and the approving review - much appreciated!

PR #452, opened 2020-08-04
Clarify moving Worldpay contracts (BYOC redux) 

jonathanglassman:
### Context

Need to clarify that you cannot move an existing direct Worldpay contract over to Government Banking.

- Clarify that you cannot move an existing direct Worldpay contract over to Government Banking
- Split out Stripe / Govt Banking vs others

### Guidance to review

Check if accurate


jonathanglassman:
> live accounts plural, maybe? ie they can create new ones?

@miriamalice which particular file and line are you referring to?

PR #456, opened 2020-08-26
Fact check a thon - Quick Start Guide

jonathanglassman:
### Context

The fact check a thon checked the accuracy of the different tech docs pages.

Clarify the following:

- 'when you've received a test account' means when you have signed up to a test account or joined the service.
- "Do not use an API key from a live account on the GOV.UK Pay admin tool. Only use a test API key." means "Do not use an API from a live account to test Pay."
- you can view test transactions in the API


kbottla (approved):
LGTM!

PR #458, opened 2020-08-26
Fact check a thon - reporting

jonathanglassman:

My service should be My services 

consider /v1/payments/{paymentId} to indicate variable more clearly (and likewise throughout) 

In text we note two things you may want to do (get a single payment, search for multiple payments), but we have sections corresponding to 3 things (<- + get payment events) 

https://docs.payments.service.gov.uk/reporting/#checking-when-your-payment-service-provider-psp-took-a-payment is a bit unclear. We store captured date for PSPs other than worldpay (stripe), and it may not always be the case that settlement will occur 2 days after capture in those cases. Might need a rethink, and also how this re4lates to payout reporting. 

The sentence 'Search results are asynchronous so the payment status may not be up-to-date.' isn't quite right. The search results aren't asynchronous. I think it would be clearer and more accurate just to say there is a delay in payment statuses in search results being updated (but better than that :)) 

The sentence 'Some of the query parameters you can use to search for payments are:' seems quite vague 

The heading 'Generate a list of payments (search payments)' seems unnecessarily verbose. Can it not just be 'Search payments'? 

Miriam:

can we add that you can download one report for all of your services in Pay as well as individual ones? 

we say that we don't show fees in a test account - but can we give users an example CSV that does include those fields? or say, email us if you want an example?

### DIFF
--- a/source/reporting/index.html.md.erb
+++ b/source/reporting/index.html.md.erb
@@ -183,9 +183,9 @@ For example:
 
 `GET /v1/payments`
 
-See the <a href="https://govukpay-api-browser.cloudapps.digital/#searchpayments" target="blank">GOV.UK Pay API browser</a> for more details
+See the <a href="https://govukpay-api-browser.cloudapps.digital/#searchpayments" target="blank">GOV.UK Pay API browser</a> for more details.
 
-Do not search payments using the API to check a payment's status during [payment flow](/payment_flow). The payment status field in search results is not up to date may not be up-to-date.
+Do not search payments using the API to check a payment's status during [payment flow](/payment_flow). The payment status field in the search results does not immediately update and so may not be up to date.
 
 ### Search criteria
 
--- a/source/reporting/index.html.md.erb
+++ b/source/reporting/index.html.md.erb
@@ -19,24 +19,21 @@ transaction information in a CSV file.
 
 You can export one file for one service's transactions, or one file for all your services' transactions.
 
-### Export one file for one service's transactions
+1. Go to [My Services](https://selfservice.payments.service.gov.uk/my-services).
 
-1. Select the account you want reporting information for in the [My Services](https://selfservice.payments.service.gov.uk/my-services) section.
+1. If you want to export one service's transactions, select the service account you want reporting information for and go to the [transactions](https://selfservice.payments.service.gov.uk/transactions) list.
 
-2. Go to the [transactions](https://selfservice.payments.service.gov.uk/transactions) list.
+    Select __View transactions for all live services__ if you want to export all your services' transactions.
 
-3. Filter transactions using the available search criteria.
+1. Filter transactions using the available search criteria if needed.
 
-4. Select __Download all transaction details as a CSV file__.
+    If you have more than 5000 transactions, you must apply a filter before the page will display your transactions.
 
-The CSV file contains all transaction data, which is much more comprehensive than the information on the __Transactions__ page. You can process CSV files using
-spreadsheet software and edit the contents you do not need.
+1. Select __Download all transaction details as a CSV file__.
 
-You can export up to 100,000 transactions.
+The CSV file contains all transaction data, which is much more comprehensive than the information on the __Transactions__ page.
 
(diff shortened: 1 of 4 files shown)


source/reporting/index.html.md.erb
```diff
@@ -111,7 +117,14 @@ For example:
...
 
-You should receive the payment in your bank account within 2 business days of `captured_date`.
+If your PSP is Stripe, you should receive the payment in your bank account within:
+
+- 2 working days of `captured_date` if your user completed their payment during the week
+- 3 working days of `captured_date` if your user completed their payment at the weekend or on a bank holiday
+
+You can confirm when payments are paid into your bank account by signing in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login) and going to __My Services__ > __View payments to your bank account__.
+
+If your PSP is Government Banking’s contracted PSP (currently Worldpay), SmartPay or ePDQ, you should receive the payment in your bank account within 2 working days of `captured_date`.
```
miriamalice, line 127: Should match the text we have in the Integrate with the API section

If your PSP is Government Banking’s contracted PSP (currently Worldpay), SmartPay or ePDQ, contact your PSP to find out your minimum payout and payment times.

source/reporting/index.html.md.erb
```diff
@@ -111,7 +117,14 @@ For example:
 }
 ```
 
-You should receive the payment in your bank account within 2 business days of `captured_date`.
+If your PSP is Stripe, you should receive the payment in your bank account within:
+
+- 2 working days of `captured_date` if your user completed their payment during the week
+- 3 working days of `captured_date` if your user completed their payment at the weekend or on a bank holiday
+
+You can confirm when payments are paid into your bank account by signing in to the [GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/login) and going to __My Services__ > __View payments to your bank account__.
```
miriamalice, line 125: Do we want to say 'when payments are paid into your bank account' or 'when transactions are paid into your bank account'? We're talking here about tracking the status of a transaction, so maybe that makes more sense? 

Also need to edit this to include the API report as well when that is released.

source/reporting/index.html.md.erb
```diff
@@ -19,24 +19,21 @@ transaction information in a CSV file.
...
+1. If you want to export one service's transactions, select the service account you want reporting information for and go to the [transactions](https://selfservice.payments.service.gov.uk/transactions) list.
 
-2. Go to the [transactions](https://selfservice.payments.service.gov.uk/transactions) list.
+    Select __View transactions for all live services__ if you want to export all your services' transactions.
 
-3. Filter transactions using the available search criteria.
+1. Filter transactions using the available search criteria if needed.
 
-4. Select __Download all transaction details as a CSV file__.
+    If you have more than 5000 transactions, you must apply a filter before the page will display your transactions.
```
kbottla, line 30: This is for CSV download link. If a service has less than 5K transactions, CSV download link is available immediately. Otherwise, a filter must be applied for CSV download link to appear.

Transactions are always displayed when services visit transactions page. Users can apply filters to refine transactions as needed

source/reporting/index.html.md.erb
jonathanglassman: @miriamalice we speak about payments in this section, so I think it's ok? Also, to check, are you referring to the https://github.com/alphagov/pay-tech-docs/pull/460 work?

miriamalice: I think payments means different things which can be a bit confusing. 
Payments into a bank account sounds like it means the total amount that is paid into a bank account - the 'payment' that you have received from Stripe.
Here I think we mean, you confirm when individual payments have been paid into your bank account.

So it might be slightly confusing but also I think it's fine. Had forgotten we use 'payments' rather than 'transactions'.

And yes, referring to 460.

jonathanglassman: ok, i'll add to the existing 460 PR to do this.

kbottla (approved):

PR #460, opened 2020-08-27
DO NOT MERGE UNTIL FEATURE IS LIVE - Payouts api

jonathanglassman:
### Context

Jira ticket: https://payments-platform.atlassian.net/browse/PP-6973

Add in content on new settlement API.

### DIFF
--- a/source/integrate_with_govuk_pay/index.html.md.erb
+++ b/source/integrate_with_govuk_pay/index.html.md.erb
@@ -82,6 +82,26 @@ If your PSP is Stripe, you can sign in to the [GOV.UK Pay admin tool](https://se
 
 If your PSP is Government Banking's contracted PSP (currently Worldpay), SmartPay or ePDQ, contact your PSP to find out your minimum payout and payment times.
 
+#### Checking when your PSP sent a payment
+
+If your PSP is Stripe, you can check when Stripe sent a payment to your bank account by making an API call to [get information about a payment](/reporting/#get-information-about-a-single-payment).
+
+Example response:
+
+```
+"settlement_summary": {
+  ...
+  "settled_date": "2016-01-22",
+}
+```
+
+where `settled_date` is either:
+
+- the date that Stripe sent the payment to your bank account as part of a group of payments (‘payout’)
+- missing if Stripe has not yet sent the payment to your bank account
+
+You can also sign in to the GOV.UK Pay admin tool to see your payout for each day, and which of your users’ payments are in each payout.
+
 ## The GOV.UK Pay API
 
 The GOV.UK Pay API offers a set of operations to conduct and report on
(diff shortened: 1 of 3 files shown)


source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -82,6 +82,26 @@ If your PSP is Stripe, you can sign in to the [GOV.UK Pay admin tool](https://se
...
+#### Checking when your PSP sent a payment
+
+If your PSP is Stripe, you can check when Stripe sent a payment to your bank account by making an API call to [get information about a payment](/reporting/#get-information-about-a-single-payment).
+
+Example response:
+
+```
+"settlement_summary": {
+  ...
+  "settled_date": "2016-01-22",
```
heathd, line 94: we should specify the timezone used for date cutoffs .. is it midnight UTC or midnight Local time?

heathd: @sfount we should ideally have a unit test specifying this behaviour .. dates have timezones too!

source/reporting/index.html.md.erb
```diff
@@ -246,9 +246,9 @@ For example, to list all the payments your PSP sent to you on 16 June 2020:
 GET /v1/payments?from_settled_date=2020-06-16T00:00:00Z&to_settled_date=2020-06-16T23:00:00Z
 ```
 
-These take inputs based on a subset of the [ISO 8601 standard](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard). For example, a valid input would be `2015-08-13T12:35:00Z`.
+These query parameters take inputs based on a subset of the [ISO 8601 standard](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard). For example, a valid input would be `2015-08-13T12:35:00Z`.
 
-If a payment has not happened yet, the payment will not be in the response.
+If a payer has not yet sent a payment, the payment will not be in the response.
```
sfount, line 251: This disambiguation is clearly useful but I believe the "payment has not happened yet" in this context refers to the Stripe payment to your bank account. 

In that case I think this might read something like 

 ```suggestion
If your PSP has not sent this payment to your bank account yet, the payment will not be in the response.
```

I'm not sure, as with the other content it's tricky because of the overloaded "payment"

source/integrate_with_govuk_pay/index.html.md.erb
sfount: Great point @heathd, currently the dates are always provided by the PSP and should always be treated as UTC. This attempts to follow the pattern set by `captured_date` which uses the same format https://govukpay-api-browser.cloudapps.digital/#tocssettlementsummary

jonathanglassman: Adding a comment to this effect.

jonathanglassman:
@sfount made changes as discussed, PTAL when you can

sfount (approved):
This is looking great 👍 a quick sense check looks like this lines up with what we'll be planning to release next week (the docs and functionality should be released together).

On one of David's points, there may be other places in the docs where clarifications around dates and date formats would help clarify but that would do well as a separate piece of work, as discussed.

PR #462, opened 2020-09-03
Add an accessibility statement for tech docs

jonathanglassman:
### Context

The deadline for publishing an accessibility statement is 23 September 2020. The GOV.UK Pay product pages already have a statement published. The content of the tech docs need an accessibility statement as well. This is separate to the statement on the tech doc template itself, which is being handled as part of a separate project.

Add an accessibility statement for tech docs.

### DIFF
--- a/source/accessibility.html.md.erb
+++ b/source/accessibility.html.md.erb
@@ -12,7 +12,7 @@ This accessibility statement applies to the GOV.UK Pay technical documentation a
 This website is run by the GOV.UK Pay team at the Government Digital Service (GDS). We want as many people as possible to be able to use this website. For example, that means you should be able to:
 
 + change colours, contrast levels and fonts
-+ zoom in up to 300% without the text spilling off the screen
++ zoom in up to 300% without problems
 + navigate most of the website using just a keyboard
 + navigate most of the website using speech recognition software
 + listen to most of the website using a screen reader (including the most recent versions of JAWS, NVDA and VoiceOver)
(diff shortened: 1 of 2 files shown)


source/accessibility.html.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+We’ve also made the website text as simple as possible to understand.
+
+[AbilityNet](https://mcmw.abilitynet.org.uk/) has advice on making your device easier to use if you have a disability.
+
+### How accessible this website is
+
+We know some parts of this website are not fully accessible.
+
+The following pages have redundant links:
```
miriamalice, line 28: what do we mean by 'redundant' here?

source/accessibility.html.md.erb
```diff
@@ -0,0 +1,120 @@
...
+
+This website is partially compliant with the [Web Content Accessibility Guidelines version 2.1](https://www.w3.org/TR/WCAG21/) AA standard, due to the non-compliances listed below.
+
+### Non-accessible content
+
+The content listed below is non-accessible for the following reasons.
+
+#### Non-compliance with the accessibility regulations
+
+The following pages have redundant links:
```
miriamalice, line 74: this seems to repeat the section above - do we need it twice?

source/accessibility.html.md.erb
```diff
@@ -0,0 +1,120 @@
...
+- inaccessible formatting in general
+- inaccessible language
+- inaccessible diagrams or images
+- lack of colour contrast for text, important graphics and controls
+- images not having meaningful alt text
+- problems when using assistive technologies such as screen readers and screen magnifiers
+
+## What we’re doing to improve accessibility
+
+We plan to look at the redundant links by [date].
```
miriamalice, line 112: do we need to commit to a date?

source/accessibility.html.md.erb
jonathanglassman: Redundant means that sequential links are the same link. It often happens when you have multiple "Contact us" links on a page, with no other links in between them. That is the case here. I've checked with the accessibility community. It's not a fail, but it does need to be called out in the statement.

source/accessibility.html.md.erb
jonathanglassman: Yes, unfortunately, that's the template. I know it's a bit ridiculous.

source/accessibility.html.md.erb
jonathanglassman: We do. I've checked with the accessibility community and there is not current guidance or advice on when this is. Maybe we could say end of 2020 or maybe end of Q1 2020?

source/accessibility.html.md.erb
miriamalice: Can we push back on the template...?

jonathanglassman: I'll ask, but I cannot make any promises :-)

jonathanglassman: Ok, so it looks like the first section is more of a summary, so i'll redraft that statement and update the PR

antoniasimmons (approved):
Happy with this statement. Just need to complete the by [date] elements.

PR #477, opened 2020-10-16
Logos can go on the right

xander-gds:
### Context
The docs currently suggest custom branding logos can only go on the left of the header, which is incorrect

Added the ability to put the logo on the right-hand side for custom branding


jonathanglassman:
LGTM

rorymalcolm (approved):
LGTM

PR #480, opened 2020-10-23
PP-6284: optionally include js commons analytics

iqbalgds:
Include all distribution code from the pay js commons analytics folder
IFF it exists (has been npm installed).
  * Importantly the build process
  should pass without the files being there, allow testing the main app
  functionality locally without analytics;
  *  the production instance will then ensure `npm install` has been
  correctly run and include the analytics client side JavaScript.
* optionally include sass required for banner
  * The tech docs gem only includes part of `govuk-frontend` that it needs.
  * The analytics consent banner require button styles. Unless this code
    becomes part of the core tech docs gem, it will need the ability to
    include these additional styles. Add an import for the button style with
    a guard against dev not having included npm dependencies locally.
* Hook show banner call into application javascripts
  * add govuk-frontend and pay-js-commons as dependencies (these should be
    optional)
  * added the show banner if no cookie consent provided line to
    application javascripts
* Add tech docs gem analytics init into javascripts
  * If we use the gem analytics ID, google analytics .js will always be
    fetched from Google no matter if consent is provided. By forking this
    and moving into our client side code, we can only run it if consent has
    been provided through the common cookie code.
* Add npm install to deploy
  * Update the README and clarify why build steps exist to make shared Pay
    assets optional.
  * Add `npm install` to the build pipeline to give production environments
    access to the client side analytics code.
* Add new tests - application.js
* Update Package.json to use pay-js-commons v3.0 which has the Cookie banner code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@
 
 .idea
 venv
+
+node_modules
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,9 +1,9 @@
 {
   "exclude": {
-    "files": null,
+    "files": "package-lock.json",
     "lines": null
   },
-  "generated_at": "2020-09-11T11:12:14Z",
+  "generated_at": "2020-10-23T08:54:01Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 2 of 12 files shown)


SandorArpa (approved):
LOKTM

PR #490, opened 2020-11-04
Improve Custom Branding content

jonathanglassman:
https://payments-platform.atlassian.net/browse/PP-6803


iqbalgds (approved):

source/optional_features/custom_branding/index.html.md.erb
```diff
@@ -19,6 +19,12 @@ customisation for the following features:
 * the background and border colour of the top banner
 * the branding of your payment confirmation email
 
+You must tell us which service you want to apply custom branding to. Custom branding applies to the test and live environments for a service.
```
miriamalice, line 22: not sure if this entirely accurate -I think we have to re-apply it to live if we do it in test? and if it's in live, it won't necessarily show up in test?

source/optional_features/custom_branding/index.html.md.erb
```diff
@@ -19,6 +19,12 @@ customisation for the following features:
...
+
+We recommend that you only enable custom branding once you have [gone live and started taking payments from real users](/switching_to_live/#go-live).
```
miriamalice, line 24: I think we sort of recommend this as it's easier for us, not necessarily because it's better for them or their testing/services. How do we handle that?

source/optional_features/custom_branding/index.html.md.erb
```diff
@@ -19,6 +19,12 @@ customisation for the following features:
...
+
+You cannot have different custom branding for payment pages and emails.
+
 ## Banner logo
 
 You should [email us](/support_contact_and_more_information/#contact-us) with an image of your desired custom banner logo. Your image should be:
```
miriamalice, line 30: are we adding an image for this?

source/optional_features/custom_branding/index.html.md.erb
```diff
@@ -35,10 +41,14 @@ directly](/support_contact_and_more_information/#contact-us).
...
+
+When you ask GOV.UK Pay to set up custom branding on your [payment confirmation emails](/payment_flow/#confirmation-email), you must tell your users how to contact you.
+
+You can do this using a reply-to email address or another way such as a contact form.
 
-When you ask GOV.UK Pay to set up custom branding on your [payment confirmation emails](/payment_flow/#confirmation-email), you should specify a reply-to email address so your users can contact you.
+If you use a reply-to email address, this email address can be a monitored email inbox for your service, or an unmonitored automated email response.
 
-This reply-to email address can be an email inbox for your service, or an automated response.
+You can customise your monitored and unmonitored email responses in the GOV.UK Pay admin tool using the `customParagraph` field in the payment receipt email template.
```
miriamalice, line 52: we need to make sure we update the team manual / templates so that this works

miriamalice: do we want to say - you should let your user know if the address is unmonitored and provide a way for them to contact your service

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: well, we could just omit the statement entirely, or keep it in unchanged and take the benefit :)

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: Kay is working on that now I believe

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: yes

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: @iqbalgds is this your understanding?

jonathanglassman:
Information from @iqbalgds: 

Ideal image dimensions would be:
300 height x 120 width pixels
This is the double density` so that it will look good on retina screens.
On the screen it will appear half that size (hope that makes sense).
A bit smaller in either dimension is fine as well.

jonathanglassman:
> Information from @iqbalgds:
> 
> Ideal image dimensions would be:
> 300 height x 120 width pixels
> This is the double density` so that it will look good on retina screens.
> On the screen it will appear half that size (hope that makes sense).
> A bit smaller in either dimension is fine as well.

@iqbalgds thanks for the info, I've added this to the content

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: Have checked with @iqbalgds and we can only apply this at the service level.

miriamalice: I think that we have to do some work to apply custom branding to live emails when we've done it with test  (we don't have to do anything on payment pages branding) - but maybe that distinction is not relevant to ours users but something we should be aware of

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: agree to remove

source/optional_features/custom_branding/index.html.md.erb
jonathanglassman: agree to address this through separate ticket on internal manual to make sure gov.uk pay team members know to check

PR #491, opened 2020-11-04
Clarify how expired payments work

jonathanglassman:
https://payments-platform.atlassian.net/browse/PP-7245

### DIFF
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -318,7 +318,7 @@ You cannot resume a payment if the `status` is `failed`, `cancelled` or `error`.
 
 When your user enters their payment information and selects __Continue__, we contact your PSP to check whether there is enough money in the user's account for the payment.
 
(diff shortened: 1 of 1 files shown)


jonathanglassman:
@sfount, miriam suggested you're the man to work with on this

sfount (commented):
This looks factually correct and clearer to me 👍  

Looking back at the Jira ticket, is there something we want to say about how the "authorisation" may be visible to user or look like "held" funds until the payment is either taken or cancelled/ expired?

How each card and bank deals with this can be different so I don't think we'd be able to say anything specific.

jonathanglassman:
> This looks factually correct and clearer to me 👍
> 
> Looking back at the Jira ticket, is there something we want to say about how the "authorisation" may be visible to user or look like "held" funds until the payment is either taken or cancelled/ expired?
> 
> How each card and bank deals with this can be different so I don't think we'd be able to say anything specific.

@sfount How about the latest amend?

sfount (approved):
Nice one, looks good 👍

PR #494, opened 2020-11-12
Remove inaccessible diagrams

jonathanglassman:
As part of the 2020 accessibility work, we should remove the following two diagrams from the tech docs as they are not accessible (use colour to denote meaning, not an svg format, alt text, not clear, not necessary):

- https://docs.payments.service.gov.uk/api_reference/#payment-status-lifecycle
- https://docs.payments.service.gov.uk/payment_flow/#overview

The payment flow overview diagram is already described in the body text so does not need any further changes.

The payment status lifecycle diagram needs to be described in the body text so this ticket will encompass that work.

### DIFF
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -77,7 +77,7 @@ A successful response includes ``status`` and ``finished`` values:
   </tr>
   <tr>
     <td> submitted </td>
-    <td> Your user has submitted payment details, and gone through 3D Secure.<br>The PSP has authorised the payment, but the user has not yet selected <b>Confirm</b>. </td>
+    <td> Your user has submitted payment details, and gone through authentication if required.<br>The PSP has authorised the payment, but the user has not yet selected <b>Confirm</b>. </td>
     <td> false </td>
   </tr>
   <tr>
@@ -87,7 +87,7 @@ A successful response includes ``status`` and ``finished`` values:
   </tr>
   <tr>
     <td> success </td>
-    <td> Your user successfully completed the payment by selecting <b>Confirm</b>. <br>The user is redirected to an appropriate payment confirmation page on your service.</td>
+    <td> Your user successfully completed the payment by selecting <b>Confirm</b>. <br>The user is redirected to an appropriate payment confirmation page.</td>
     <td> true </td>
   </tr>
   <tr>
--- a/source/api_reference/index.html.md.erb
+++ b/source/api_reference/index.html.md.erb
@@ -50,10 +50,8 @@ Authorization: Bearer <YOUR-API-KEY-HERE>
 
 ## Payment status lifecycle
 
-The following diagram gives an overview of the payment status lifecycle and the
-possible outcomes, excluding for delayed capture payments:
+The following table summarises the payment status lifecycle and the possible outcomes, excluding for delayed capture payments.
 
-<%= image_tag "/images/payment-states.svg", { :alt => '' } %>
 
 You can check the status of a payment using the <a
 href="https://govukpay-api-browser.cloudapps.digital/#get-a-payment"
@@ -69,7 +67,7 @@ A successful response includes ``status`` and ``finished`` values:
   </tr>
   <tr>
     <td> created </td>
-    <td> Payment created. Your user has not yet visited <code>next_url</code>. </td>
+    <td> Payment created using the API. Your user has not yet visited <code>next_url</code>. </td>
     <td> false </td>
   </tr>
   <tr>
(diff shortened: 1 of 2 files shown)


source/api_reference/index.html.md.erb
```diff
@@ -79,22 +77,22 @@ A successful response includes ``status`` and ``finished`` values:
   </tr>
   <tr>
     <td> submitted </td>
-    <td> Your user has submitted payment details but has not yet selected <b>Confirm</b>. </td>
+    <td> Your user has submitted payment details, and gone through 3D Secure.<br>The PSP has authorised the payment, but the user has not yet selected <b>Confirm</b>. </td>
```
miriamalice, line 80: do we need to say 'if applicable' or something for 3dS?

source/api_reference/index.html.md.erb
```diff
@@ -79,22 +77,22 @@ A successful response includes ``status`` and ``finished`` values:
...
   <tr>
     <td> capturable </td>
-    <td> The payment is a [delayed capture](/delayed_capture/#delay-taking-a-payment), and your user has submitted payment details and selected <b>Confirm</b>. </td>
+    <td> The payment is a <a href="/delayed_capture/#delay-taking-a-payment" target="blank">delayed capture</a>, and your user has submitted payment details and selected <b>Confirm</b>. </td>
     <td> false </td>
   </tr>
   <tr>
     <td> success </td>
-    <td> Your user successfully completed the payment. </td>
+    <td> Your user successfully completed the payment by selecting <b>Confirm</b>. <br>The user is redirected to an appropriate payment confirmation page on your service.</td>
```
miriamalice, line 90: do we need to say what would happen with payment links (ie they got the confirmation page that we host?)

source/api_reference/index.html.md.erb
```diff
@@ -104,7 +102,7 @@ A successful response includes ``status`` and ``finished`` values:
   </tr>
   <tr>
     <td> error </td>
-    <td> Something went wrong with GOV.UK Pay or the underlying payment service provider. <a href="/support_contact_and_more_information/#contact-us">Contact us.</a> </td>
+    <td> Something went wrong with GOV.UK Pay or the underlying payment service provider. Payment fails safely with no money taken. <a href="/support_contact_and_more_information/#contact-us">Contact us.</a> </td>
```
miriamalice, line 105: what do we show the user here? and error message as well?

source/payment_flow/index.html.md.erb
```diff
@@ -13,25 +13,13 @@ There's a different process for [taking Direct Debit payments](/direct_debit/#ta
 
 ## Overview
 
-The following diagram outlines the payment process for a service that has a GOV.UK Pay
-integration:
+The following payment process is for a service that is integrated with the GOV.UK Pay API.
```
miriamalice, line 16: at some point it might be good to say - there is a different process for payment links and provide a link with details on that? recognise it's a dicsussion as to whether that sits in the product pages or the tech docs.

source/payment_flow/index.html.md.erb
```diff
@@ -13,25 +13,13 @@ There's a different process for [taking Direct Debit payments](/direct_debit/#ta
...
-GOV.UK Pay.
-
-Your user enters their payment details (for example payment card
-details and billing address) on the GOV.UK Pay pages. GOV.UK Pay verifies the
-payment with the underlying PSP.
-
-After the transaction reaches a final state, your user is then redirected
-back to your service.
+1. When your user needs to make a payment, your service makes an API call to create a new payment.
+1. Your service then redirect your use to the payment pages hosted on GOV.UK Pay.
```
miriamalice, line 19: redirects?

jonathanglassman: whoops

source/payment_flow/index.html.md.erb
jonathanglassman: Directly above this on the page at https://docs.payments.service.gov.uk/payment_flow/#how-gov-uk-pay-works is a warning with a link to the payment links page in the tech docs: `Warning - This does not apply to your users who use the payment links functionality`. Is that sufficient do you think? I don't want to repeat the same link again so soon.

source/api_reference/index.html.md.erb
jonathanglassman: How about "Strong Customer Authentication" instead as that's the term we're using now?

miriamalice: we should use that, yes, good point. but also it only applies to online payments not phone payments, so not all transactions with have SCA.

source/payment_flow/index.html.md.erb
miriamalice: yeah that's fine, sorry hadnt seen that. would be good to think about payment links specific content at some point too.

source/api_reference/index.html.md.erb
jonathanglassman: I think we just remove "on your service" here to account for both situations. If we explain both situations, the table cell gets really long.

source/api_reference/index.html.md.erb
jonathanglassman: I don't know. Maybe we should put a link to the HTTP status code content below this table?

jonathanglassman: @sfount @kbottla can you advise?

kbottla: Users will see below, if there is a technical error
Screenshot 2020-11-18 at 17 15 43

We display this hint in selfservice for Error status

```
The was an error with the payment. This could be due to a configuration problem between the service and the payment provider. Contact the GOV.UK Pay team with the payment ID, quoting error code P0050.
```

jonathanglassman: thank you!

jonathanglassman:
@miriamalice I've incorporated Krishna's latest information into this PR. Are you happy with the content?

miriamalice (approved):

PR #507, opened 2020-12-03
PP-7416 Add ‘Select Turn on 3DS Flex’ to 3DS Flex instructions

alexbishop1:
### Context
Services now need to press a button to enable 3DS Flex after they’ve entered their 3DS Flex credentials. They’ll be on the page with the button after they’ve entered the credentials and the button will not appear if the credentials have not been entered.

Add ‘Select **Turn on 3DS Flex**’ to list of instructions.

### Guidance to review
image


jonathanglassman (approved):
LGTM

PR #509, opened 2020-12-04
Differentiate between payment links and integrating with the API

jonathanglassman:
As per https://payments-platform.atlassian.net/browse/PP-7410

The current Go Live docs are heavily focused on the API journey, e.g. it says that a user needs to generate an API key. 

This may be confusing for users who are going live with payment links, as not everything is applicable to them.

This PR is to make it clearer what is required to go live with payment links compared to the API, and also apply this to the rest of the tech docs.

### DIFF
--- a/source/making_payments/index.html.md.erb
+++ b/source/making_payments/index.html.md.erb
@@ -7,6 +7,8 @@ weight: 50
 
 # Take a payment
 
+This section tells you how to take a payment using the GOV.UK Pay API.
+
 When your user makes a payment, your service should supply a `reference`. If
 possible, this should be a unique identifier for the payment. If your
 organisation already has an existing identifier for payments, you can use it
@@ -16,8 +18,6 @@ Your service will also need to supply a
 [`return_url`](#choose-the-return-url-and-match-your-users-to-payments). This
 is a URL that your service hosts for your user to return to, after their
 payment journey on GOV.UK Pay ends.
-
-<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
 
 ## Creating a payment
 
--- a/source/making_payments/index.html.md.erb
+++ b/source/making_payments/index.html.md.erb
@@ -17,6 +17,8 @@ Your service will also need to supply a
 is a URL that your service hosts for your user to return to, after their
 payment journey on GOV.UK Pay ends.
 
+<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
+
 ## Creating a payment
 
 The call to create a payment with the GOV.UK Pay API is:
--- a/source/optional_features/index.html.md.erb
+++ b/source/optional_features/index.html.md.erb
@@ -11,5 +11,3 @@ You can customise your payments pages with GOV.UK Pay in the following ways:
 * [use Welsh](/optional_features/welsh_language/#use-welsh-on-your-payment-pages) on payment pages
 * [use your own payment failure pages](/optional_features/use_your_own_error_pages)
 * [prefill fields](/optional_features/prefill_user_details/) on the **Enter card details** page
-
-You can customise your payments pages regardless of whether you [integrate with the GOV.UK Pay API](/integrate_with_govuk_pay/#integrate-with-the-gov-uk-pay-api) or use [payment links](https://www.payments.service.gov.uk/govuk-payment-pages/).
--- a/source/optional_features/index.html.md.erb
+++ b/source/optional_features/index.html.md.erb
@@ -11,3 +11,5 @@ You can customise your payments pages with GOV.UK Pay in the following ways:
 * [use Welsh](/optional_features/welsh_language/#use-welsh-on-your-payment-pages) on payment pages
 * [use your own payment failure pages](/optional_features/use_your_own_error_pages)
 * [prefill fields](/optional_features/prefill_user_details/) on the **Enter card details** page
+
+You can customise your payments pages regardless of whether you [integrate with the GOV.UK Pay API](/integrate_with_govuk_pay/#integrate-with-the-gov-uk-pay-api) or use [payment links](https://www.payments.service.gov.uk/govuk-payment-pages/).
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -11,8 +11,6 @@ This section outlines how your service will interact with GOV.UK Pay after you i
 
 There's a different process for [taking Direct Debit payments](/direct_debit/#take-a-direct-debit-payment).
 
-<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
-
 ## Overview
 
 The following payment process is for a service that is integrated with the GOV.UK Pay API.
--- a/source/payment_flow/index.html.md.erb
+++ b/source/payment_flow/index.html.md.erb
@@ -11,7 +11,7 @@ This section outlines how your service will interact with GOV.UK Pay after you i
 
 There's a different process for [taking Direct Debit payments](/direct_debit/#take-a-direct-debit-payment).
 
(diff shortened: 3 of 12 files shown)


source/optional_features/index.html.md.erb
```diff
@@ -11,3 +11,5 @@ You can customise your payments pages with GOV.UK Pay in the following ways:
 * [use Welsh](/optional_features/welsh_language/#use-welsh-on-your-payment-pages) on payment pages
 * [use your own payment failure pages](/optional_features/use_your_own_error_pages)
 * [prefill fields](/optional_features/prefill_user_details/) on the **Enter card details** page
+
```
miriamalice, line 14: some of these customisations are available with payment links and some aren't. 
- custom branding can be done on payment links
- welsh is available on payment links (it's in the set up)
- failure pages is API
- prefill fields is API

I'd be inclined not to change this page. it should be clear from the details of each one if it's in the API call or not.

source/payment_flow/index.html.md.erb
```diff
@@ -11,7 +11,7 @@ This section outlines how your service will interact with GOV.UK Pay after you i
 
 There's a different process for [taking Direct Debit payments](/direct_debit/#take-a-direct-debit-payment).
 
-<%= warning_text('This does not apply to your users who use the <a href="https://www.payments.service.gov.uk/govuk-payment-pages/">payment links</a> functionality') %>
+<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
```
miriamalice, line 14: this is also a slightly different call than the regular create a payment API.
since we're not really doing on DD at the moment I'd be inclined leave this page as is.

source/quick_start_guide/index.html.md.erb
```diff
@@ -26,8 +26,9 @@ If you want to build a technical integration between your service and the
 GOV.UK Pay API, your service team should have the necessary skills. You can
 refer to the [GOV.UK Service
 Manual](https://www.gov.uk/service-manual/the-team/what-each-role-does-in-service-team#roles-your-team-must-have)
-for more information. This does not apply if you only use [payment
-links](https://www.payments.service.gov.uk/govuk-payment-pages/).
+for more information.
+
+<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
```
miriamalice, line 31: some of this text does apply and some doesn't.

- all users should read about Pay and create an account (and probs read the service manual).
- only API users need to look at the swagger file

Maybe we introduce earlier the idea of the two types of integration, ie decide if you want to integrate with the API or use payment links. (instead of the warning, which maybe feels a but aggressive to the poor payment link users!)

Perhaps there's some content frm our new Get Started page that is relevant here?

source/refunding_payments/index.html.md.erb
```diff
@@ -13,6 +13,8 @@ You can also refund a payment by signing into the [GOV.UK Pay admin tool](https:
 
 Your user will receive the refund in the same account they originally paid from. You cannot refund a payment to another card or bank account.
 
+You can refund payments regardless of whether you [integrate with the GOV.UK Pay API](/integrate_with_govuk_pay/#integrate-with-the-gov-uk-pay-api) or use [payment links](https://www.payments.service.gov.uk/govuk-payment-pages/).
```
miriamalice, line 16: I don't know that we need to say this. Maybe just calling out where it is different for payment links is sufficient?

source/security/index.html.md.erb
```diff
@@ -17,6 +17,8 @@ Please do not disclose the suspected breach publicly until it has been fixed.
 
 GOV.UK Pay allows you to create as many API keys as you want.
 
+This only applies if you [integrate with the GOV.UK Pay API](/integrate_with_govuk_pay/#integrate-with-the-gov-uk-pay-api).
```
miriamalice, line 20: A user can technically create an API key on a payment links service but they don't need to. I wonder if we just leave this section as is?

source/set_up_3dsecure/index.html.md.erb
```diff
@@ -15,6 +15,8 @@ If you only take payments within the UK, you must implement SCA by 14 September
 
 To implement SCA, you must set up 3D Secure 2 (3DS2) or 3D Secure 1 (3DS1) payment authentication.
 
+You must implement SCA regardless of whether you [integrate with the GOV.UK Pay API](/integrate_with_govuk_pay/#integrate-with-the-gov-uk-pay-api) or use [payment links](https://www.payments.service.gov.uk/govuk-payment-pages/).
```
miriamalice, line 18: Again, not sure it's necessary to call out when they are the same

source/switching_to_live/index.html.md.erb
```diff
@@ -94,21 +94,46 @@ You need to make the following changes to your live account if you made them to
...
+
+1. After you've confirmed the payment, check you received a [confirmation email](/payment_flow/#confirmation-email).
 
-6. Check the payment appears in [your transactions](https://selfservice.payments.service.gov.uk/transactions) in the GOV.UK Pay admin tool.
+1. Check the payment appears in [your transactions](https://selfservice.payments.service.gov.uk/transactions) in the GOV.UK Pay admin tool.
 
-7. Select the payment and check you can refund it - it may take up to 20 minutes for the __Refund payment__ button to appear.
+1. Select the payment and check you can refund it. It may take up to 20 minutes for the __Refund payment__ button to appear.
 
 If your payment or refund fails, check:
```
miriamalice, line 138: Do we also need to add something to this section about testing Pay?
https://docs.payments.service.gov.uk/testing_govuk_pay/#testing-the-whole-user-journey ? We link to it from the Go Live section which should be applicable to both payment link and API users, so maybe the testing section needs to say you can test a payment link...?

source/optional_features/index.html.md.erb
jonathanglassman: ok will change back

source/payment_flow/index.html.md.erb
jonathanglassman: ok will change back

source/quick_start_guide/index.html.md.erb
jonathanglassman: we could add in the great content from Justin - the stuff from the "Decide how you want to take payments" section of the get started content?

source/refunding_payments/index.html.md.erb
jonathanglassman: ok will remove

source/security/index.html.md.erb
jonathanglassman: ok will remove

source/set_up_3dsecure/index.html.md.erb
jonathanglassman: ok

source/switching_to_live/index.html.md.erb
jonathanglassman: ok i'll draft something

jonathanglassman:
> we've seen that some service teams build a srivce using a payment lnk when they should really invest in an integration. should we offer any further differentiation on this page (ie you can automate your service with our integration) or would we expect people to have done that research themselves on the product pages first? happy to leave as is for now, but something to bear in mind - i think this makes the payment links sound super appealing with no downside but it's definitely not right for every service.

agree, i'll take out the positive blurb

source/making_payments/index.html.md.erb
```diff
@@ -17,6 +17,8 @@ Your service will also need to supply a
 is a URL that your service hosts for your user to return to, after their
 payment journey on GOV.UK Pay ends.
 
+<%= warning_text('This does not apply if you only use <a href="https://www.payments.service.gov.uk/govuk-payment-pages/" target="_blank">payment links</a>.') %>
```
miriamalice, line 20: should we add an intro to this section to say that this is about taking a payment using the API? (rather than having a warning about it not applying to payment links)? but also fine either way. Just feels like we jump into this page without a lot of context

jonathanglassman: agree, have amended

miriamalice (approved):

PR #537, opened 2021-01-18
PP-7528 Add deadline for teams' own Worldpay accounts

m-green:
### Context
We're missing content for users to tell them that you cannot use your own Worldpay contract after 31 July 2021.

Add the content to the relevant [Go live section](https://docs.payments.service.gov.uk/switching_to_live/#2-connect-your-live-account-to-your-psp) and reorder the content so the options are clearer. 

### Guidance to review


antoniasimmons (approved):
Yep, think this covers the missing gap. Technically the wording around how you can connect ePDQ and Smartpay on the basis you've done this previously is also true for Worldpay direct contract use but it's a bit nuanced. As we discussed I am happy this is the simplest and clearest approach for now. Cheers for sorting.

PR #539, opened 2021-01-25
PP-6441 Add asking us about storing records of phone payments

m-green:
### Context
If teams use Worldpay as their PSP and a third-party provider that processes phone payments, they can ask us about storing records about phone payments on GOV.UK Pay.

- Add a page on asking us about storing records of phone payments
- Restructure so we have a 'Taking payments by phone' section that contains both MOTO payments and storing records of phone payments 

### To do before merging
- [ ] Add redirects to reflect tweaked URLs
- [ ] Update links to MOTO page from other sections
 
### Guidance to review


miriamalice (commented):
lgtm but I have the least context of this of the three reviewers!

heathd:
I think this feature is quite niche. It's also quite a lot of work for our team to support organisations using this feature so I'm not keen on advertising it. I would prefer to have a page in the documentation which describes the scenario where this feature might be used without actually mentioning the existence of the feature itself and invite people to get in touch with us.

heathd (changes requested):
This feature is very niche, and I'm not keen on advertising it because it creates quite a lot of burden on our team to manage the integration work required. Therefore I would prefer that we do not put this in our public documentation.

m-green:
Thanks David - definitely happy to avoid having content about this feature if it's a niche user need.

I wasn't sure what you meant by the alternative page though - can you elaborate a little?

heathd:
Instead of:

> # Take payments by phone
> 
> You can:
> 
> - [take payments over the phone](moto-payments.html) (‘MOTO’ payments)
> - ask us about [storing records about phone payments](store-records-about-phone-payments.html)

Something like:

> # Take payments by phone
> 
> You can:
> 
> - [take payments by having your staff key in card payment details in a GOV.UK Pay](moto-payments.html) you will be responsible for PCI compliance for the way that your staff handle the user's card details. This feature is sometimes called ‘MOTO’ payments
> - take payments by [outsourcing the operation of your call centre to a specialist telephony supplier](using-an-outsourced-telephony-provider-with-govuk-pay.html).

And then a page such as:

> # Using an outsourced telephony provider with GOV.UK Pay
> 
> Many services outsource the operation of their call centre to a specialist supplier who can manage PCI compliance. These suppliers usually offer ways to keep credit card numbers secure by automating the card number entry using telephone touch tones (sometimes called "DTMF suppression"). GOV.UK Pay does not currently offer a way to allow such suppliers to process payments, however we do have limited support to allow payments processed outside of GOV.UK Pay to be stored in the platform for reporting purposes. Currently this is only supported if you are using Worldpay.
>
> If you're interested to know more about this, please contact us

m-green:
Thanks David. I think the suggested content might be tipping into a higher-level need that’s a better fit for the product pages.

However I’ve created an iterated and shortened version below to bring it all into GOV.UK style and/or the language already agreed in the NHS/DWP Google Doc. Let me know what you think. (Please comment rather than rewriting so we can avoid steering too far away from style etc).

I haven’t included PCI compliance considerations as it seemed unnecessary to the user need of "I need to use GOV.UK Pay to process phone payments handled by a third-party" - if a team is using a call centre, it seems like they’d already know about PCI compliance and DTMF suppression.

I lost 'limited support' because it would probably raise questions about what that means (which means teams would be unsure and probably just contact us anyway). Happy to add it in if we can find words that are a little more specific, to help teams decide whether to contact us.

# Take payments by phone

You can take payments by phone by [getting your staff to fill in GOV.UK Pay payment pages with your users’ details](https://docs.payments.service.gov.uk/moto_payments/#take-a-payment-over-the-phone-moto-payments) - also known as Mail Order / Telephone Order (MOTO) payments.

If a third-party provider processes your phone payments, you cannot use GOV.UK Pay to process those payments. However if your PSP is Worldpay, you can [contact us] about whether you can store records about phone payments in your GOV.UK Pay account, so you can report on both your online and phone payments.

PR #540, opened 2021-01-27
PP-6479 Clarify support page and make it more consistent with product support page

m-green:
### Context
A fact check on the Support page highlighted some areas for clarification.

I've borrowed content from the [product support page](https://www.payments.service.gov.uk/support/) to make it clearer how to contact us in an emergency.

I've also taken the opportunity to:

- restructure the page with fewer headings so it's easier to understand (this is especially important given that this page will be used in emergencies), and closer to the product support page
- make the P1 (critical) description match the product support page, to avoid confusion through inconsistency
- fix ambiguities and style issues across the page as a whole
- make improvements to the incident table - the style is now closer to the GOV.UK style already used by the GDS Way incident table

I haven't added clarification that the contact email isn't for end users. We've researched and tried a similar approach on other products but it doesn't seem to reduce accidental emails from end users who should be contacting the relevant department.

### Guidance to review

### DIFF
--- a/source/support_contact_and_more_information/index.html.md.erb
+++ b/source/support_contact_and_more_information/index.html.md.erb
@@ -9,45 +9,32 @@ weight: 220
 
 You can get notifications about GOV.UK Pay’s status to your email, phone, RSS reader or website, by subscribing on our [status page](https://payments.statuspage.io/#).
 
-## Incident severity
+After you contact us, the GOV.UK Pay team will contact you using the Zendesk ticketing system.
 
-GOV.UK Pay incident severities are defined as follows:
+## Report an incident
 
-<div style="height:1px;font-size:1px;">&nbsp;</div>
+How quickly we respond depends on the type of your incident.
 
-| Classification | AKA | Example | Initial response time | Update Time |
+| Classification | Type | Example | Initial response time | Update frequency |
 | :--- | :--- | :--- | :--- | :--- |
-| P1 | Critical Incident | complete outage; substantial degradation of service; significant security issue; significant availability issue  | 30 minutes (at any time) | 1hr |
-| P2 | Major Incident | elevated error rate; mostly up; noticeably degraded service; upstream vulnerabilities; complete component failure | 60 minutes (in waking hours) | 2hrs |
-| P3 | Significant | Users experiencing intermittent or degraded service due to platform issue | 1 business day or next working day | 2 business days |
-| P4 | Minor | Component failure that is not immediately service impacting | 2 business days | weekly |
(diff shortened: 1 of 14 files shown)


source/support_contact_and_more_information/index.html.md.erb
```diff
@@ -9,45 +9,32 @@ weight: 220
 
 You can get notifications about GOV.UK Pay’s status to your email, phone, RSS reader or website, by subscribing on our [status page](https://payments.statuspage.io/#).
 
-## Incident severity
+After you contact us, the GOV.UK Pay team will contact you using the Zendesk ticketing system.
```
miriamalice, line 12: Should we include a general intro here to our support process? e.g. Our office hours are 9.30am to 5.30pm, Monday to Friday. You can email us at govuk-pay-support@digital.cabinet-office.gov.uk. We’ll reply within 2 working days. The team are available out of hours in an emergency. Otherwise I wonder if this support page feels a bt like it just talks about incidents?

source/support_contact_and_more_information/index.html.md.erb
```diff
@@ -9,45 +9,32 @@ weight: 220
...
 
-GOV.UK Pay incident severities are defined as follows:
+## Report an incident
```
miriamalice, line 14: should we tell people where to find the emergency contact info? it's in the email we send when a service went live (and it's in some of the MoUs).

m-green: Did you see I added it further down the page?

source/support_contact_and_more_information/index.html.md.erb
m-green: Ah good point - I can add a bit of this to the '## General feedback' section, as that's where we're meeting this need. Do we reply to general questions and feedback within 2 days?

source/support_contact_and_more_information/index.html.md.erb
miriamalice: oh sorry yes missed this!

source/support_contact_and_more_information/index.html.md.erb
miriamalice: Yes, 2 working days for general questions. 1 working day response for go live requests.

m-green: @miriamalice Thanks - now added in https://github.com/alphagov/pay-tech-docs/pull/540/commits/b619876347215fad393db9fb270068475b0551a1

kbottla (approved):
LGTM!

miriamalice (approved):

PR #541, opened 2021-01-29
PP-7719 Update 3DS page with fact check amends

m-green:
### Context
A fact check on the 3DS page have highlighted some things that need fixing. 

Factual amends.

### Guidance to review
@alexbishop1 - adding you as just wanted to check if we have the steps right under '### Set up 3DS2 (3DS Flex)'?

### DIFF
--- a/source/set_up_3dsecure/index.html.md.erb
+++ b/source/set_up_3dsecure/index.html.md.erb
@@ -15,7 +15,7 @@ If you only take payments within the UK, you must implement SCA by 14 September 
 
 To implement SCA, you must set up 3D Secure 2 (3DS2) or 3D Secure 1 (3DS1) payment authentication.
 
-Payment service providers (PSPs) recommend setting up 3DS2, because 3DS1 will be retired in 2022.
+Payment service providers (PSPs) recommend setting up 3DS2, because they'll start retiring 3DS1 in 2022.
 
 How you set up 3DS2 or 3DS1 payment authentication depends on which Payment Service Provider (PSP) you’re using.
 
--- a/source/set_up_3dsecure/index.html.md.erb
+++ b/source/set_up_3dsecure/index.html.md.erb
@@ -1,6 +1,6 @@
 ---
 title: Set up Strong Customer Authentication (3DS1 and 3DS2)
-last_reviewed_on: 2020-08-28
+last_reviewed_on: 2021-01-29
 review_in: 6 months
 weight: 170
 ---
@@ -9,11 +9,13 @@ weight: 170
 
 Strong Customer Authentication (SCA) makes online payments more secure for users and reduces fraud. SCA is part of the Payment Service Directive 2 (PSD2).
 
-If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA by 31 December 2020.
+If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA when you [go live](/switching_to_live/#go-live).
 
 If you only take payments within the UK, you must implement SCA by 14 September 2021.
 
 To implement SCA, you must set up 3D Secure 2 (3DS2) or 3D Secure 1 (3DS1) payment authentication.
+
+Payment service providers (PSPs) recommend setting up 3DS2, because 3DS1 will be retired in 2022.
 
 How you set up 3DS2 or 3DS1 payment authentication depends on which Payment Service Provider (PSP) you’re using.
 
@@ -36,9 +38,10 @@ Worldpay does not apply any SCA exemptions to transactions.
 1. In the GOV.UK Pay admin tool, select the account you want to set up on the __My services__ page.
 1. In __Settings__, make sure 3D Secure is set to __On__.
 1. In __Your PSP__, enter your 3DS Flex credentials.
+1. Select __Save credentials__.
 1. Select __Turn on 3DS Flex__.
 
-<%= warning_text('If your credentials are wrong, you are not able to take payments.') %>
+If you turn off 3DS Flex later, your account will use 3DS1 by default.
 
 ### Check 3DS1 is enabled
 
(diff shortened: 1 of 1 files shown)


source/set_up_3dsecure/index.html.md.erb
```diff
@@ -9,12 +9,14 @@ weight: 170
...
 Strong Customer Authentication (SCA) makes online payments more secure for users and reduces fraud. SCA is part of the Payment Service Directive 2 (PSD2).
 
-If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA by 31 December 2020.
+If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA when you [go live](/switching_to_live/#go-live).
 
 If you only take payments within the UK, you must implement SCA by 14 September 2021.
 
 To implement SCA, you must set up 3D Secure 2 (3DS2) or 3D Secure 1 (3DS1) payment authentication.
 
+Payment service providers (PSPs) recommend setting up 3DS2, because 3DS1 will be retired in 2022.
```
alexbishop1, line 18: My understanding is that Mastercard plans to “sunset” (Worldpay’s words) 3DS1 in October 2022 but I’m not sure about Visa and the like. So we might want to weaken this slightly and say that 3DS1 will begin to be retired in 2022. We could weave this in with the extra fees for 3DS1 (some of which are already in place) and say something like 3DS1 is becoming more expensive and will begin to be retired altogether in 2022.

source/set_up_3dsecure/index.html.md.erb
```diff
@@ -36,9 +38,10 @@ Worldpay does not apply any SCA exemptions to transactions.
 1. In the GOV.UK Pay admin tool, select the account you want to set up on the __My services__ page.
 1. In __Settings__, make sure 3D Secure is set to __On__.
 1. In __Your PSP__, enter your 3DS Flex credentials.
+1. Select __Save credentials__.
```
alexbishop1, line 41: Whoops, should have updated this myself when I tinkered with this page!

alexbishop1 (changes requested):
Looks good apart from one statement we might want to make less unequivocal.

source/set_up_3dsecure/index.html.md.erb
stevenjmesser: I missed it too, to be fair.

source/set_up_3dsecure/index.html.md.erb
m-green: Ah thanks @alexbishop1. We don't say anything about 3DS1 fees at the moment - @stevenjmesser, should we add something on that?

stevenjmesser: Fees are between the service and their PSP, but it would be helpful to add what we know about the bigger picture. Are these tech docs the best place for that though? Perhaps we should iterate the 3DS page on the product pages.

m-green: Makes sense - shall we go with this as is in the tech docs in that case (with Alex's suggested tweak to "begin to be retired")?

stevenjmesser: I reckon so, yeah.

m-green: @stevenjmesser @alexbishop1 Tweaked in https://github.com/alphagov/pay-tech-docs/pull/541/commits/3630dacdeba3e4986220613f04acc88a372020ce and ready for a green tick if you're both happy

alexbishop1:
Yes, 3DS is run by the banks and card schemes, though the PSPs in front of them.

source/set_up_3dsecure/index.html.md.erb
```diff
@@ -9,12 +9,14 @@ weight: 170
...
 Strong Customer Authentication (SCA) makes online payments more secure for users and reduces fraud. SCA is part of the Payment Service Directive 2 (PSD2).
 
-If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA by 31 December 2020.
+If you take payments from the European Economic Area (EEA), defined as the European Union, Norway, Iceland and Liechtenstein, you must implement SCA when you [go live](/switching_to_live/#go-live).
 
 If you only take payments within the UK, you must implement SCA by 14 September 2021.
 
 To implement SCA, you must set up 3D Secure 2 (3DS2) or 3D Secure 1 (3DS1) payment authentication.
 
+Payment service providers (PSPs) recommend setting up 3DS2, because they'll start retiring 3DS1 in 2022.
```
alexbishop1, line 18: I think the second clause should be more along the lines of “because 3DS1 will start to be retired in 2022” to reflect it’s the card schemes driving this.

alexbishop1 (changes requested):
Fine apart from one thing.

source/set_up_3dsecure/index.html.md.erb
m-green: Unfortunately "3DS will start to be retired in 2022" is passive voice - can we find an active version by adding the responsible party? From the conversation above - would "because card companies will start to retire..." be ok?

alexbishop1: Microsoft Word always used to tell me I use the passive voice too much. That and my sentences are too long.

I think the term to use is probably “card schemes” and adding in the examples of Visa and Mastercard is probably necessary because a lot of people won’t know what card schemes are.

m-green: Thanks - updated in https://github.com/alphagov/pay-tech-docs/pull/541/commits/0f54015cde7a8a077f06846065a3cb8624cc69e2.

m-green: p.s. passive vs active is hard! It's so easy to use it accidentally. 2iers often spot it in my own content where I've missed it.

alexbishop1 (approved):

m-green:
@stevenjmesser Are you happy to green tick this one? If so it's all good to go.

stevenjmesser:
> @stevenjmesser Are you happy to green tick this one? If so it's all good to go.

So sorry, @m-green, I saw that Alex had approved it but didn't think to check if one had been requested from me!

This looks good, thanks so much!

stevenjmesser (approved):

PR #542, opened 2021-02-01
PP-7721 Replace Direct Debit content with research contact

m-green:
### Context
Replacing the content about signing up for the Direct Debit pilot with contacting us about taking part in research. 

### Guidance to review


stevenjmesser (approved):
Looks good to me!

PR #547, opened 2021-02-04
PP-6802 Improve accounts page

m-green:
### Context
We've received requests from users who are confused about or need more help setting up multiple accounts.

This improves the [How accounts work](https://docs.payments.service.gov.uk/account_structure/#how-accounts-work) tech docs page to better meet user needs.

I’ve aimed to:

- lead with the need for each type of setup, so users can more easily pick the approach that matches what they’re trying/need to do
- use language like “send your users…” and “send payments…” to make it easier to understand the flow of payments through accounts
- make the content more task-based (so it uses headings like ‘Decide where to…’ and now includes how to actually set up accounts in the different ways)

We’re a bit stuck with how the admin tool uses “accounts” and “services” to mean different things, but I’ve aimed to be consistent as possible here using:

- “online service” or “your service” for team services
- “GOV.UK Pay service”
- “PSP account”

I’ve removed the diagrams for now because I think they might be helping rather than hindering. I’m probably 50:50 on the diagrams though, so would be good to test without diagrams and see if there’s a need. 

I’ve been through some related Zendesk tickets to check that we’re using users’ language as far as possible, and we’re meeting needs around setting up multiple accounts.

I’ve removed some content that we felt didn’t belong on this page (including moving bank payment descriptions to the 'Take a payment' page), and tweaked content that we felt wasn’t clearly explained (although the ‘changing bank accounts’ content could be improved further I think).

### Guidance to review

### DIFF
--- a/source/account_structure/index.html.md.erb
+++ b/source/account_structure/index.html.md.erb
@@ -1,54 +1,59 @@
 ---
-title: Structuring your accounts
+title: Setting up multiple services
 last_reviewed_on: 2021-02-03
 review_in: 6 months
 weight: 40
 ---
 
-# Deciding whether to combine accounts and payments
+# Setting up multiple services
   
-You should set up your GOV.UK Pay services and payment service provider (PSP) accounts so you can more easily report on and reconcile your payments, and manage your services. 
+If you have more than one online service, you can set up your GOV.UK Pay account and payment service provider (PSP) accounts so it's easier to manage your payments and services.
 
-GOV.UK Pay services are free, and there's no limit to how many you can create. Your PSP may charge you for each PSP account you have.
+## Decide where to send users
 
-## Decide whether to combine your users' payments 
+You can send your users to make payments with either:
 
-Each GOV.UK Pay service you create can take payments for either:
+- a different GOV.UK Pay service for each of your services
+- a combined GOV.UK Pay service
 
-- one of your services
-- 2 or more of your services combined
+### Send your users to different GOV.UK Pay services
 
-### Take payments for one service
+Send your users to a different GOV.UK Pay service for each of your services if you need to do any of the following:
 
-If a GOV.UK Pay service takes payments for only one of your services, you can:
+- give each service its own name on the GOV.UK Pay payment pages
+- report on each service's payments separately
+- send each service's payments to different bank accounts
+- give your staff members access to each service separately in the GOV.UK Pay admin tool
 
-- give your service its own name on GOV.UK Pay's payment pages
-- report on your service's payments separately
-- send your service's payments to a dedicated bank account
-- give your staff members access to your service separately in the GOV.UK Pay admin tool
+To set up your services like this, [create a new GOV.UK Pay service](https://selfservice.payments.service.gov.uk/my-services/create) for each of your online services.
+
+GOV.UK Pay services are free, and there's no limit to how many you can create.
 
 To change a service's name, go to the [**My services** page in the GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/my-services).
 
-### Take payments for 2 or more services combined
+### Send your users to a combined GOV.UK Pay service
 
-If a GOV.UK Pay service takes payments for 2 or more of your services combined:
+Send your users to a combined GOV.UK Pay service if you need to:
 
-- you can more easily report on all the services' payments together
-- you can more easily change a setting across all the services, for example which card types you accept
-- you may need to use a different [reference number](/making_payments/#reference) or add [custom metadata](/custom_metadata/#add-custom-metadata) for each of your services, so you can tell which service each payment comes from
-- you cannot give the services different names
-- you cannot send payments to different bank accounts
+- report on all the services' payments together
+- change settings easily across all the services, for example which card types you accept
 
-To set up your services like this, use the same API key to create a payment in all of your services that you want to combine.
+You will not be able to give the services different names or send payments to different bank accounts.
 
-## Decide whether to combine your bank payments
+You may need to use a different [reference number](/making_payments/#reference) or add [custom metadata](/custom_metadata/#add-custom-metadata) for each of your services, so you can tell which service each payment comes from.
 
-To help you reconcile payments, each PSP account you create can receive money from either:
+To set up a group of your services like this, use the same API key or payment link for all the services in the group.
 
-- one GOV.UK Pay service
-- 2 or more GOV.UK Pay services combined - unless you use Stripe
+## Decide where to send payments
 
-If you want one PSP account to recieve money from 2 or more GOV.UK Pay services, add the same merchant code to each GOV.UK Pay service when you [go live](/switching_to_live/#go-live).
+To help you reconcile payments, each GOV.UK Pay service can send payments to a:
+
+- separate PSP account and bank account
+- combined PSP account and back account - unless you use Stripe
+
+To set up a combined PSP account, add the same PSP merchant code to each GOV.UK Pay service when you [go live](/switching_to_live/#go-live).
+
+If you send payments to separate PSP accounts, your PSP may charge you for each PSP account you have.
 
 You cannot send money from one GOV.UK Pay service to different PSP accounts.
 
--- a/source/account_structure/index.html.md.erb
+++ b/source/account_structure/index.html.md.erb
@@ -1,92 +1,63 @@
 ---
-title: How accounts work
-last_reviewed_on: 2021-01-08
+title: Structuring your accounts
+last_reviewed_on: 2021-02-03
 review_in: 6 months
 weight: 40
 ---
 
-# How accounts work
+# Deciding whether to combine accounts and payments
+  
+You should set up your GOV.UK Pay services and payment service provider (PSP) accounts so you can more easily report on and reconcile your payments, and manage your services. 
 
-GOV.UK Pay does not limit how many services you can integrate with it and how many accounts you can have with us.
-GOV.UK Pay test and live accounts are free to create and maintain but your payment service provider may charge you for the underlying accounts.
+GOV.UK Pay services are free, and there's no limit to how many you can create. Your PSP may charge you for each PSP account you have.
 
-## Account structure examples
+## Decide whether to combine your users' payments 
 
-If you do integrate multiple services, you can treat them separately or as a
-combined single service within GOV.UK Pay and your payment service provider (PSP).
-This gives 3 possible account configurations.
+Each GOV.UK Pay service you create can take payments for either:
 
-As an example, take 2 services called ‘parking permits’ and ‘parking fines’:
+- one of your services
+- 2 or more of your services combined
 
-### Type 1 (GOV.UK Pay separate, PSP separate)
+### Take payments for one service
 
-<%= image_tag "/images/accountstructure_1.svg", { :alt => '' } %>
+If a GOV.UK Pay service takes payments for only one of your services, you can:
 
-The diagram show two services, ‘Parking permits’ and ‘Parking fines’, that are considered separate services by
-both GOV.UK Pay and the PSP.
+- give your service its own name on GOV.UK Pay's payment pages
+- report on your service's payments separately
+- send your service's payments to a dedicated bank account
+- give your staff members access to your service separately in the GOV.UK Pay admin tool
 
-If you use this account structure, you can:
+To change a service's name, go to the [**My services** page in the GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/my-services).
 
-- give access for each service to different staff members
-- optionally use different service names on the GOV.UK Pay payment pages
-- send each service's payments to a different bank account
-- report separately on each of your services
+### Take payments for 2 or more services combined
 
-### Type 2 (GOV.UK Pay combined, PSP combined)
+If a GOV.UK Pay service takes payments for 2 or more of your services combined:
 
-<%= image_tag "/images/accountstructure_2.svg", { :alt => '' } %>
+- you can more easily report on all the services' payments together
+- you can more easily change a setting across all the services, for example which card types you accept
+- you may need to use a different [reference number](/making_payments/#reference) or add [custom metadata](/custom_metadata/#add-custom-metadata) for each of your services, so you can tell which service each payment comes from
+- you cannot give the services different names
+- you cannot send payments to different bank accounts
 
-The diagram shows two services, ‘Parking permits’ and ‘Parking fines’, that are considered a single combined
-service by GOV.UK Pay and the PSP.
+To set up your services like this, use the same API key to create a payment in all of your services that you want to combine.
 
-You may want to use this configuration if you want to simplify the administration of multiple services in GOV.UK Pay. For example, if you do not need reporting at an individual service level.
+## Decide whether to combine your bank payments
 
-The transactions from both services will appear in the same transactions list and CSV file. You may find it useful to use reference numbers to separate payments clearly, for example “parking_fine_1234” and “parking_permit_1234”.
+To help you reconcile payments, each PSP account you create can receive money from either:
 
-In this configuration, payments to services must go to the same bank account because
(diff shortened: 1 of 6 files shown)


source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
+- report on each service's payments separately
+- send each service's payments to different bank accounts
+- give your staff members access to each service separately in the GOV.UK Pay admin tool
 
-The diagram show two services, ‘Parking permits’ and ‘Parking fines’, that are considered separate services by
-both GOV.UK Pay and the PSP.
+To set up your services like this, [create a new GOV.UK Pay service](https://selfservice.payments.service.gov.uk/my-services/create) for each of your online services.
 
-If you use this account structure, you can:
+GOV.UK Pay services are free, and there's no limit to how many you can create.
```
miriamalice, line 30: Maybe we need to say, if you have lots of services set up in Pay, you'll need to manage settings on each one. Basically want to remind them that there is an overhead to managing lots of services individually - needing to add/remove users when they leave, needing to change settings on each one if they change what card types they accept. sort of referenced in the section below but I have a feeling it's slightly overlooked as a long term consideration

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-- give access for each service to different staff members
-- optionally use different service names on the GOV.UK Pay payment pages
-- send each service's payments to a different bank account
-- report separately on each of your services
+To change a service's name, go to the [**My services** page in the GOV.UK Pay admin tool](https://selfservice.payments.service.gov.uk/my-services).
```
miriamalice, line 32: happy to keep this here as a reminder but not sure it's directly related?

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-### Type 2 (GOV.UK Pay combined, PSP combined)
+### Send your users to a combined GOV.UK Pay service
 
-<%= image_tag "/images/accountstructure_2.svg", { :alt => '' } %>
+Send your users to a combined GOV.UK Pay service if you need to:
 
-The diagram shows two services, ‘Parking permits’ and ‘Parking fines’, that are considered a single combined
-service by GOV.UK Pay and the PSP.
+- report on all the services' payments together
```
miriamalice, line 38: you can do this even if you have lots of services - you'd just use the all transactions view, so not sure this is a need. But would be good to be clear on the different levels of reporting (and the availability of this all transactions report).

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
+- change settings easily across all the services, for example which card types you accept
```
miriamalice, line 39: or add and remove users

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-You may want to use this configuration if you want to simplify the administration of multiple services in GOV.UK Pay. For example, if you do not need reporting at an individual service level.
+You will not be able to give the services different names or send payments to different bank accounts.
 
-The transactions from both services will appear in the same transactions list and CSV file. You may find it useful to use reference numbers to separate payments clearly, for example “parking_fine_1234” and “parking_permit_1234”.
+You may need to use a different [reference number](/making_payments/#reference) or add [custom metadata](/custom_metadata/#add-custom-metadata) for each of your services, so you can tell which service each payment comes from.
```
miriamalice, line 43: You can use:
different reference
different description
add metadata

to tell which service each payment comes from or add information for reporting/reconciliation

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-In this configuration, payments to services must go to the same bank account because
-they are a single service in the PSP. It is not possible to support multiple PSP accounts in a single GOV.UK Pay service.
+To set up a group of your services like this, use the same API key or payment link for all the services in the group.
```
miriamalice, line 45: you can set up multiple payment links in one service. this means you can change the settings in each one (like reference number field name, description, reporting columns). but they will all be on the same CSV download within that service. a user would have to get all the data for that service and then filter by description / reporting column to separate out different transaction types.

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-### Type 3 (GOV.UK Pay separate, PSP combined)
+## Decide where to send payments
 
-<%= image_tag "/images/accountstructure_3.svg", { :alt => '' } %>
+To help you reconcile payments, each GOV.UK Pay service can send payments to a:
```
miriamalice, line 49: I think we can make this a bit more specific about the options with Stripe and Worldpay? (i'm relatively happy to ignore epdq and smartpay since we're not onboarding new people, so current users will likely have figured this out by now)

If you use Stripe, you will get a separate payout for each service in Pay. You can choose to have all payouts go into the same bank account, or provide a different bank account for each one during the go live process. You can set the description that will appear on your bank statement next to each payout.

If you use Worldpay, you add a Merchant code to each service. You will receive one payout for each merchant code you use. You can choose to add the same Merchant code to all Pay services, or you can use different Merchant codes. (I don't think Worldpay charge for merchant codes under the gov banking contract)

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
+### Send your users to different GOV.UK Pay services
 
-### Type 1 (GOV.UK Pay separate, PSP separate)
+Send your users to a different GOV.UK Pay service for each of your services if you need to do any of the following:
 
-<%= image_tag "/images/accountstructure_1.svg", { :alt => '' } %>
+- give each service its own name on the GOV.UK Pay payment pages
+- report on each service's payments separately
+- send each service's payments to different bank accounts
+- give your staff members access to each service separately in the GOV.UK Pay admin tool
```
miriamalice, line 27: maybe we also need something about being able to manage settings individually on different accounts? e.g. what card types are accepted, the custom paragraph in the confirmation email. Do we have a link to those in the docs or would the Settings page in the admin tool be the best place to see all of those?

source/account_structure/index.html.md.erb
miriamalice: There's also something we could say about the granularity of services... (thugh maybe this gets too complicate for round 1 of improvements, or is best shown in an example...)

An LA may want a 'registrar office' service, and within that they can have API and/or payment links for each type of certificate - marriage, birth etc. The can use our reporting features to differentiate between payments, and everyone in the registrar office can see all the transactions.  They then set up a 'Parking' service, and a 'Garden Waste' service etc, reflecting different team areas.

But some LAs are setting up a 'birth certificate' service, and a 'marriage certificate' service. My guess is they all have the same users, the same settings (though actually, that's something to check!) In which case I think they are doing this for reporting reasons...? but there's an overhead - everytime someone leaves the team they need to remove them for 10 services instead of 1. etc. (maybe that's minimal, but feels inefficient). \

Anyways that is maybe too much info, but I do think knowing where to draw the line in granularity is an issue for services...

source/account_structure/index.html.md.erb
miriamalice: oh and we should add that if you are using Worldpay and need to take MOTO payments, you need a separate service for that (as you need a separate merchant code for ECOMM vs. MOTO https://docs.payments.service.gov.uk/moto_payments/#take-a-payment-over-the-phone-moto-payments

source/making_payments/index.html.md.erb
```diff
@@ -78,6 +78,11 @@ The format is variable. If you have an existing format, you can keep using it if
 
 The description must be no longer than 255 characters, and must not contain URLs.
 
+The `description` does not appear on your users' bank statements. To change your users' bank statements, you can either:
```
miriamalice, line 81: I'm not sure this quite flows here - since the rest of the page is so much about the API. I think we can move it into the reporting /reconciliation bit...?

miriamalice (commented):
Hopefully these comments make sense - very happy to talk through any and all of them! Thanks for such a quick turnaround on this!

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 ---
 
-# How accounts work
+# Setting up multiple services
+  
+If you have more than one online service, you can set up your GOV.UK Pay account and payment service provider (PSP) accounts so it's easier to manage your payments and services.
 
-GOV.UK Pay does not limit how many services you can integrate with it and how many accounts you can have with us.
-GOV.UK Pay test and live accounts are free to create and maintain but your payment service provider may charge you for the underlying accounts.
+## Decide where to send users
```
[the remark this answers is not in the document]
m-green: That makes sense - as discussed in 81e419d I've rejigged everything into a new set of pages that breaks this down into distinct (but interlinked) tasks. So the 'where to send users' ideas is gone now.

source/account_structure/index.html.md.erb
```diff
@@ -1,92 +1,68 @@
...
 
-## Account structure examples
+You can send your users to make payments with either:
 
-If you do integrate multiple services, you can treat them separately or as a
-combined single service within GOV.UK Pay and your payment service provider (PSP).
-This gives 3 possible account configurations.
+- a different GOV.UK Pay service for each of your services
```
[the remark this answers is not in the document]
m-green: Hopefully the way things are broken down in 81e419d means that we can guide users through their particular scenario. I'd like to start without specific scenarios if possible because I'm not sure they'll be particularly helpful - from the support tickets I'm not sure each team's situation maps nicely to others, so specific examples may lead to even more mental overhead. Happy to add later if testing/evidence shows we need them though.

source/account_structure/index.html.md.erb
m-green: On the first point: see if this looks better in 81e419d. Now it's broken down a bit more, it should hopefully be clearer to users what the advantages/disadvantages of each approach are.

On the second point: I agree that granularity of services may over-complicate things at this stage, so fully going into this is probably out of scope. However I've gone partway there by attempting to add a bit about using different API keys/payment links but I'm not sure I have the details right, so please do suggest improvements.

source/account_structure/index.html.md.erb
m-green: Cool - I've removed this. Think it's a clear journey in the admin tool now we're making My Services the default page.

source/account_structure/index.html.md.erb
m-green: Ah good point - I've removed this. There are now links to the Reporting page too so all the info about this should be clear.

source/account_structure/index.html.md.erb
m-green: Ah thanks - I've added this. I've also added using different API keys / payments links but may have the details wrong?

source/account_structure/index.html.md.erb
m-green: As mentioned in another comment -  I've added using different API keys / payments links to the list of ways you can distinguish between services, but may have the details wrong?

source/account_structure/index.html.md.erb
m-green: Thanks for this - added!

source/account_structure/index.html.md.erb
m-green: On the first point: good point. I've added this. I don't think we need to list the settings - they're clear from the admin tool.

On the second point: I'm not sure we could get into MOTO here. It's not going to be a criteria for many users that affects their decisions, and we give specific instructions for creating a separate account "if you do not have one already" on the MOTO page itself.

source/making_payments/index.html.md.erb
m-green: Makes sense, thanks - I've moved it.

m-green:
Hi @miriamalice - thanks for the chat yesterday - new iteration incoming!

I've responded to your comments above but this is quite a significant overhaul from the last iteration, so a couple of things to note.

- In the 'Using a combined bank account' section, the steps include creating a separate GOV.UK Pay account for each service. This seemed to make more sense because it makes the process clearer (use the same bank account/merchant code across all GOV.UK Pay services). I've linked to the alternate approach (ie using a combined GOV.UK Pay account)
- I've moved the page down the navigation order because it now feels like it fits much more closely with Integrating (because of the clear relationship to reconciling) and Go live (because of the relationship with account settings). 

Hope it all makes sense - happy to chat through!

source/account_structure/set_up_where_services_link_to/index.html.md.erb
```diff
@@ -0,0 +1,42 @@
...
+
+You can use one GOV.UK Pay service to take payments for a group of your services, if you need to easily manage the following across the whole group:
+
+- your staff members' access in the GOV.UK Pay admin tool
+- your account settings, for example which card types you accept
+
+1. [Create one GOV.UK Pay service](https://selfservice.payments.service.gov.uk/my-services/create).
+2. Use API keys or payment links from the GOV.UK Pay service in all of the services in your group.
+
+You will not be able to give your services different names or send payments to different bank accounts.
```
miriamalice, line 24: You can change the description the user sees when making the payment. 

(I don't think we make ti very clear in the docs or product pages where all the infrmation you enter ends up on the payment page - ie service name is at the top, description is at the side. not for this PR but something to consider elsewhere....)

m-green:
Thanks @miriamalice - updates in https://github.com/alphagov/pay-tech-docs/pull/547/commits/e094221f833b98e3d95c1494c6d6e047db444535.

I tried to fit in the detail about payouts but it doesn't fit logically into the way the content is set up (for example - we can't say 'you'll get a payout per merchant code' where we talk about all services having the same merchant code). So I've added a link to the payout content at the top of the page instead - hope that's ok. (This will also help avoid duplication.)

source/account_structure/set_up_where_payments_go/index.html.md.erb
```diff
@@ -0,0 +1,58 @@
...
+---
+
+# Set up which bank accounts your payments go to
+
+To help you reconcile payments, you can send payments from your services to:
+
+- the same bank account
+- different bank accounts
+
+GOV.UK Pay services are free, and there's no limit to how many you can create. Your payment service provider (PSP) may charge you for each PSP account you have, unless you use GOV.UK Pay's PSP.
```
miriamalice, line 15: The phrasing here sort of makes the two things (does your PSP charge you / does Stripe charge you) feel more dependent on each other? I'd be tempted to separate it out and just say - there's no charge for GOV.UK Pay's PSP.

source/account_structure/set_up_where_services_link_to/index.html.md.erb
```diff
@@ -0,0 +1,45 @@
...
+If you need to know which service each payment came from when you [report on payments](/reporting), you can:
+
+- add different [reference numbers](/making_payments/#reference)
+- add different [descriptions](/making_payments/#description)
+- use [custom metadata](/custom_metadata/#add-custom-metadata)
+- use different [payment links](https://www.payments.service.gov.uk/govuk-payment-pages/)
+
+## Link your services to different GOV.UK Pay services
+
+You should [create a separate GOV.UK Pay service](https://selfservice.payments.service.gov.uk/my-services/create) for each of your services if you need to do any of the following:
```
miriamalice, line 38: i wonder if we do need to add a link to MOTO in here. def not essential but I wonder if some people might come here before investigating the MOTO page?

source/account_structure/set_up_where_payments_go/index.html.md.erb
m-green: Thank you - updated in 160e014

m-green:
Hi @miriamalice - I thought of a way to add the payouts clarification without unpicking the structure. Take a look at https://github.com/alphagov/pay-tech-docs/pull/547/commits/da4efdf1965b2bfc4716c38d5659b884b2876a66 for the changes

miriamalice (approved):

SteveEvansGDS (approved):

PR #555, opened 2021-02-23
PP-7727 review and fact check go live pages

SandorArpa:
### Context
- the pages contained in this commit are all correct, links are working and fact checked


m-green (approved):
Thank you Sandor! 👍

PR #557, opened 2021-02-24
PP-7727 Fix to go live page after fact check Feb 2021

m-green:
### Context
A fact check highlighted an issue with the content on the go live page.

Update the journey to PSP settings. (I've handed off to the UI after reaching **Settings** because it's intuitive from there. Describing the steps in too much detail is more work for the user - and increases the risk the content will fall out of sync with UI changes.)

### Guidance to review


SandorArpa (approved):
LOKTM

PR #562, opened 2021-03-11
PP-7866 Add non-US Stripe test cards

m-green:
### Context
Our [table of payment cards for Stripe test accounts](https://docs.payments.service.gov.uk/testing_govuk_pay/#if-you-39-re-using-a-test-stripe-account) is missing non-US cards. 

Add non-US cards to the table, and guidance on the transaction fee varying depending on country.

### Guidance to review


kbottla (approved):
LGTM!

PR #579, opened 2021-04-14
PP-7744 - Adds information about agreeing to Pay's terms and conditions on the 'Go Live' page

NathanD-GDS:
### Context
When requesting a Live account for Pay, users need to provide details of their organisation, tell us which Payment Service Provider they'll be using, and agree to our terms and conditions. Currently the 'Go live' page in the Pay docs only lists the first two steps. This change adds the third step (agreeing to terms and conditions).


NathanD-GDS:
Closing PR until publishing freeze is over.

stephencdaly (approved):

PR #587, opened 2021-04-22
PP-8045 Tech docs - review and fact check (March 2021)

SandorArpa:
### Context
- review and fact check tech docs March 2021 edition
- no need for any changes
- updated the last reviewed dates.


nimalank7 (approved):
LGTM!

PR #588, opened 2021-04-27
PP-8015 Fix and update the version of pay-js-commons used.

iqbalgds:
- Old version of 'pay-js-commons' used the wrong version number of 3.1.0.
- Update the pay-js-commons to the latest version with the correct version number
  3.0.12


SandorArpa (approved):
LOKTM

PR #589, opened 2021-04-28
PP-7591 - Adds a link to API standards iteration guidance

NathanD-GDS:
### Context
There have been [several enquiries](https://payments-platform.atlassian.net/browse/PP-7591) about whether we follow the API standards and how likely it is that GOV.UK Pay integrations could break when we update the API. 

This PR adds a link to the section of the API standard around iteration. It also updates some outdated wording left over from when the Pay API was in beta.

### Guidance to review
My questions are:

- Is just linking to the 'Iteration' section of the standards enough?
- Do we need to provide any more information around what adherence to the API standards entails?

### DIFF
--- a/source/versioning/index.html.md.erb
+++ b/source/versioning/index.html.md.erb
@@ -7,7 +7,10 @@ weight: 190
 
 # Stay up to date
 
-The GOV.UK Pay API follows the [Semantic Versioning 2.0.0 system](https://semver.org/).
+The GOV.UK Pay API follows the [API data and technical standards](https://www.gov.uk/guidance/gds-api-technical-and-data-standards#when-iterating-your-api)
+to protect your services from disruption when we update.
+
+We follow the [Semantic Versioning 2.0.0 system](https://semver.org/).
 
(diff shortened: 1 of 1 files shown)


alexbishop1 (approved):
This looks good.

We might be able to change the last paragraph about the support period time frames because I think in reality we’d work with services to figure out mutually-agreeable time frames rather than impose a deadline on them. But we can change this text later.

source/versioning/index.html.md.erb
```diff
@@ -7,7 +7,11 @@ weight: 190
 
 # Stay up to date
 
-The GOV.UK Pay API follows the [Semantic Versioning 2.0.0 system](https://semver.org/).
+The GOV.UK Pay API follows the Central Digital and Data Office's [API data and technical standards](https://www.gov.uk/guidance/gds-api-technical-and-data-standards#when-iterating-your-api).
```
jonathanglassman, line 10: Change link to `https://www.gov.uk/guidance/gds-api-technical-and-data-standards` as it follows the entirety of the standard.

PR #606, opened 2021-05-21
Clarifies date formatting for refund dates and settled dates.

NathanD-GDS:
### Context
In a previous Pay tech docs fact check, Steven F. pointed out that ISO 8601 has different levels of precision when specifying times and dates. We currently do not outline the level of precision that the Pay API requires.

This PR explains the level of date and time precision required for the `from_date`, `to_date`, `from_settled_date`, and `to_settled_date` query parameters require. It also changes the formatting of an example API call to match other call examples in the documentation.

### Guidance to review
For pre-i:

- I'm not sure if I describe the level of accuracy required as well as I could in line 294

### DIFF
--- a/source/making_payments/index.html.md.erb
+++ b/source/making_payments/index.html.md.erb
@@ -305,7 +305,7 @@ that’s not `null`, you can cancel the payment. For example:
 
 ```
 "cancel": {
-  "href": "https://publicapi.payments.service.gov.uk/v1/payments/{paymentId}/cancel",
+  "href": "https://publicapi.payments.service.gov.uk/v1/payments/{PAYMENT_ID}/cancel",
   "method": "POST"
 }
 ```
--- a/source/making_payments/index.html.md.erb
+++ b/source/making_payments/index.html.md.erb
@@ -305,7 +305,7 @@ that’s not `null`, you can cancel the payment. For example:
 
 ```
 "cancel": {
-  "href": "https://publicapi.payments.service.gov.uk/v1/payments/{PAYMENT_ID}/cancel",
+  "href": "https://publicapi.payments.service.gov.uk/v1/payments/hu20sqlact5260q2nanm0q8u93/cancel",
   "method": "POST"
 }
 ```
--- a/source/refunding_payments/index.html.md.erb
+++ b/source/refunding_payments/index.html.md.erb
@@ -291,7 +291,7 @@ All dates are in Coordinated Universal Time (UTC).
 
 Dates must be in [ISO
 8601 format](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard).
-You must include a date value, a time value expressed in at least hours and minutes, and a time zone designation. For example, `2019-09-20T13:30Z`.
+You must include a date value, a time value precise to at least hours and minutes, and a time zone designation. For example, `2019-09-20T13:30Z`.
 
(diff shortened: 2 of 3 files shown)


source/making_payments/index.html.md.erb
```diff
@@ -305,7 +305,7 @@ that’s not `null`, you can cancel the payment. For example:
 
 ```
 "cancel": {
-  "href": "https://publicapi.payments.service.gov.uk/v1/payments/{paymentId}/cancel",
+  "href": "https://publicapi.payments.service.gov.uk/v1/payments/{PAYMENT_ID}/cancel",
```
m-green, line 308: Possibly out of scope, but I wonder if this example should have an actual example of a payment id (eg `hu20sqlact5260q2nanm0q8u93`) rather than the placeholder? Because a placeholder suggests something the user needs to replace. That would be consistent with other response examples further up the page.

source/refunding_payments/index.html.md.erb
```diff
@@ -287,8 +287,11 @@ You can use the following query parameters to filter refunds by date:
 * `from_date`, which is the start date to search, inclusive
 * `to_date`, which is the end date to search, exclusive
 
+All dates are in Coordinated Universal Time (UTC).
+
 Dates must be in [ISO
-8601 format](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard). For example `2019-09-20T13:30:00Z`.
+8601 format](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard).
```
m-green, line 293: Just an idea, but I wonder whether bullet format might make this a bit clearer. I think the descriptions could be made a little more concise too (for example "a time value precise to at least hours and minutes" could be shortened to "include hours and minutes")

For example:

Dates must be in [ISO 8601 format] and include a:

- date
- hours and minutes - seconds are optional
- time zone

For example `2019-09-20T13:30Z`.

source/reporting/index.html.md.erb
```diff
@@ -395,9 +395,11 @@ You can use the following query parameters to filter payments by date:
 * `from_date` - the start date for payments to be searched, inclusive
 * `to_date` - the end date for payments to be searched, exclusive
 
-These take inputs based on a subset of [the ISO
-8601 standard](https://www.gov.uk/government/publications/open-standards-for-government/date-times-and-time-stamps-standard).
-For example, a valid input would be `2015-08-13T12:35:00Z`.
+All dates are in Coordinated Universal Time (UTC).
+
+Dates must be in [ISO
```
m-green, line 400: See my comment on the refunds page - suggest the same here.

m-green (commented):
Nice clarifications! Just a couple of comments from me.

source/refunding_payments/index.html.md.erb
```diff
@@ -301,11 +309,10 @@ All dates are in Coordinated Universal Time (UTC).
 
 For example, to list all the refunds your PSP took on 16 June 2020:
 
-```
-GET /v1/refunds?from_settled_date=2020-06-16&to_settled_date=2020-06-16
-```
+`GET /v1/refunds?from_settled_date=2020-06-16&to_settled_date=2020-06-16`
```
jonathanglassman, line 312: How come this is no longer in code format?

stephencdaly (approved):

PR #613, opened 2021-06-23
Updated the review date

stevenjmesser:
Reviewed the content and it looks good.

However, next time we update this, we could release telephone payment links as a public beta / live feature.


NathanD-GDS (approved):
I've added it to my to-do list and will update this page in the next couple of weeks.

PR #620, opened 2021-07-12
PP-8220 - Switching PSP guidance

NathanD-GDS:
### Context
Services can no switch their payment service provider (PSP) to Worldpay or Stripe.

Added a new section and 3 new pages to the documentation. A top-level page with information about switching in general, then a page of guidance each for Stripe and Worldpay.

### DIFF
--- a/source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
+++ b/source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
@@ -109,7 +109,7 @@ You have now switched PSP to Stripe. Stripe will process new payments.
 
 After you switch to Stripe, you can still refund old payments through your old PSP.
 
-GOV.UK Pay will maintain the old PSP integration for up to 3 months after the PSP took the last payment to allow your service to process refunds through your old PSP.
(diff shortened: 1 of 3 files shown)


source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
```diff
@@ -0,0 +1,116 @@
...
+
+## Switch to Stripe
+
+You need to repeat the following process for each service you want to switch.
+
+### Provide your details and test the integration with Stripe
+
+To test the integration between Stripe and GOV.UK Pay, you need to make a live payment. This step makes sure you have integrated your GOV.UK Pay and Stripe accounts correctly.
+
+The live payment is £2.00 and is refundable (minus the transaction fee).
```
stephencdaly, line 56: Could this be a tiny bit confusing as to whether the person making their payment only gets back an amount minus the transaction fee, rather than them getting back the full amount but the service's new Stripe account being debited the transaction fee?

stephencdaly (commented):
All looks good to me! Just one minor question

stephencdaly (approved):

PR #640, opened 2021-08-18
Corrected an inconsistent reference to Pay

NathanD-GDS:
### Context
Pay should be referred to as 'GOV.UK Pay'.

Change one reference of 'Pay' to 'GOV.UK Pay'


nimalank7 (approved):
LGTM!

PR #649, opened 2021-09-13
Clarified payment link functionality after switching PSP.

NathanD-GDS:
FCDO were unsure about whether payment links would be affected by switching PSP. This PR clarifies the behaviour under the 'What the switch affects' heading.


xander-gds (approved):
LGTM

PR #670, opened 2021-11-03
PP-8519 Google Analytics - set domain on Google cookie

iqbalgds:
- Fix bug with the domain not getting set correctly on the
  Google Analytics GA cookie.
- Explicitly set the domain when running `ga('create', ...)`


SandorArpa (approved):
LOKTM

PR #695, opened 2022-01-14
Adds new Stripe test card

NathanD-GDS:
### Context
New Stripe fees are being introduced in January 2022. From April 2022, Pay services will be responsible for these fees. There is not currently a way to test 3DS fees using test cards in our documentation.

Adds a new Stripe test card that guarantees a 3DS check so that services can test their integrations with the new fees.

### DIFF
--- a/source/testing_govuk_pay/index.html.md.erb
+++ b/source/testing_govuk_pay/index.html.md.erb
@@ -95,7 +95,7 @@ The [PSP transaction fee](/reporting/#psp-fees) depends on which country the tes
 |Successful payment |5105105105105100| Mastercard | Debit | US |
 |Successful payment |5200828282828210| Mastercard | Debit | US |
 |Successful payment |371449635398431| American Express | Credit | US |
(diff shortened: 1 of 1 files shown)


source/testing_govuk_pay/index.html.md.erb
```diff
@@ -95,6 +95,7 @@ The [PSP transaction fee](/reporting/#psp-fees) depends on which country the tes
 |Successful payment |5105105105105100| Mastercard | Debit | US |
 |Successful payment |5200828282828210| Mastercard | Debit | US |
 |Successful payment |371449635398431| American Express | Credit | US |
+|Requires 3DS check |4000002500003155| Visa | Credit | France
```
stephencdaly, line 98: It's missing a `|` at the end of the line. Not sure if this is required for it to format correctly

NathanD-GDS: @stephencdaly Good spot! I think I've added that now.

PR #701, opened 2022-01-21
Improved the description of what a service should do with return URL users

NathanD-GDS:
### Context
Some services have been trying to use the search payments API call to check the `status` of payments after users reach the `return_url`. This was providing inaccurate `status` information.

Improve the description of what services should do when their users are directed to their `return_url`.


jonathanglassman (approved):

alexbishop1 (commented):
This looks good and I like the link to the specific page about getting a single payment.

I think we might want to add something saying that they should not use any other API URL because it might not have up information.

I can see a service reading this and thinking, for example, “I don’t have easy access to the payment ID at this point in my code but I do have the reference, so I’m just going to call GET /v1/payments?reference=PAYMENT_REFERENCE` instead.”

PR #706, opened 2022-02-07
PP-9155 update verify your PSP docs - take a test payment

SandorArpa:
### Context
- Service cannot take a test payment until Stripe has confirmed their details. Next step button is not visible if the process is still ongoing.

### DIFF
--- a/source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
+++ b/source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
@@ -117,7 +117,7 @@ The live payment is £2.00 and is refundable. Your new Stripe account will be de
 
 1. Select **Make a live payment to test your Stripe PSP** - you will need a debit or credit card for this step.
 
-1. Select **Continue to live payment**.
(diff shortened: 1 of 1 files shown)


source/switch_payment_service_provider/switch_psp_to_stripe/index.html.md.erb
```diff
@@ -118,6 +118,7 @@ The live payment is £2.00 and is refundable. Your new Stripe account will be de
 1. Select **Make a live payment to test your Stripe PSP** - you will need a debit or credit card for this step.
 
 1. Select **Continue to live payment**.
+<%= warning_text('You might have to wait while Stripe verifies your details. If `Continue to live payment` button is not visible, please try later.') %>
```
NathanD-GDS, line 121: I don't think this needs to be a warning. I would just add it as a new sentence in the same step. So line 120 should be:

`1. Select **Continue to live payment**. If **Continue to live payment** is not visible, Stripe are still verifying your details. Try this step again later.`

NathanD-GDS (changes requested):
Just a little tweak. Thank you for spotting this and adding it!

SandorArpa:
> Just a little tweak. Thank you for spotting this and adding it!

It's a new feature that we just implemented, not something that was missed earlier :)

PR #736, opened 2022-05-27
Uses markdown with <nobr> tags, rather than pure HTML, to make a table.

NathanD-GDS:
### Context
To prevent wrapping issues with table entries, we've previously manually created HTML tables in the tech docs. HTML tables are trickier to edit and stick out because the rest of the docs are entirely in Markdown. 

This PR changes the last HTML table to markdown and uses <nobr> tags to prevent wrapping. The content is identical.


iqbalgds (approved):
LGTM - I never knew about the <nobr> tag - so good to know!

PR #742, opened 2022-06-10
PP-9738 review and fact check 06/22

SandorArpa:
### Context
- Daniel the Manual Spaniel has alerted us that these pages are overdue for a review.


NathanD-GDS (approved):
LGTM

PR #752, opened 2022-07-14
Adds guidance to prefill payment link details 

NathanD-GDS:
# This PR will be merged when the feature goes live

### Context
Users can now add `amount` and `reference` query parameters to a payment link. This prefills the payment amount and payment reference fields when the user opens the link.

Typically we haven't documented payment link features as payment links as the 'non-technical' way to take payments. However, given that this feature currently requires manually editing of the URL, we think it's technical enough for a docs page.

This PR adds a new page to the Pay documentation explaining this functionality.

### DIFF
--- a/source/prefill_payment_links/index.html.md.erb
+++ b/source/prefill_payment_links/index.html.md.erb
@@ -1,0 +1,40 @@
+---
+title: Prefill payment reference and amount when using a payment link
+last_reviewed_on: 2022-07-21
+review_in: 6 months
+weight: 104
+---
+
+# Prefill payment reference and amount when using a payment link
+
+You can prefill the payment reference and amount when you send a payment link to your users. When the user opens your payment link, the **Total to pay** and reference fields will already be completed.
+
+Prefill these details by adding `reference` and `amount` information to your payment link URL using query parameters. You can use one or both parameters.
+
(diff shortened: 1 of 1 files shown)


source/prefill_payment_links/index.html.md.erb
```diff
@@ -0,0 +1,40 @@
+---
+title: Prefill payment reference and amount when using a payment link
+last_reviewed_on: 2022-07-21
+review_in: 6 months
+weight: 104
+---
+
+# Prefill payment reference and amount when using a payment link
+
+You can prefill the payment reference and amount when you send a payment link to your users. When the user opens your payment link, the **Total to pay** and reference fields will already be completed.
```
iqbalgds, line 10: Do we want to bold both `total to pay` and `reference`?
Just wondering why just one item is in bold?

iqbalgds (changes requested):

source/prefill_payment_links/index.html.md.erb
NathanD-GDS: Good spot! The reason for that is because services can name the reference field, so we don't know what that field will be called. Total to pay is the same for every service, though, as far as I'm aware.

iqbalgds: Sound good - will approve now.

iqbalgds (approved):

PR #771, opened 2022-09-01
Fixes endpoint formatting, paragraph order, and return_url behaviour

NathanD-GDS:


source/recurring_payments/index.html.md.erb
```diff
@@ -197,10 +199,10 @@ Your request should look something like this:
 
 The API returns a standard create payment response. Successful responses include the `agreement_id` you sent and `authorisation_mode` is `agreement`.
 
-To check if a payment authorised successfully, wait a few moments and get information about it through the API (`GET v1/payments/{PAYMENT_ID}`).
-
 GOV.UK Pay uses a token representing the user’s saved payment details from the agreement to automatically take the payment.
 
+To check if a payment authorised successfully, wait a few moments and get information about it through the API (`GET /v1/payments/{PAYMENT_ID}`).
```
sfount, line 204: 🤔 you can also `"use [GOV.UK Pay’s webhooks to receive an automatic update](/recurring_payments/#use-webhooks-to-receive-automatic-event-updates)"` for when the payment succeeded as well. I wonder there is a general statement about checking the status of the payment, which can be done either receiving automatic updates or waiting some time and calling the GET payments endpoint.

sfount (approved):
Looks good! Small question on the difference between succeeded/ failed

source/recurring_payments/index.html.md.erb
NathanD-GDS: That's a really good shout. I'll update the PR with that change!

NathanD-GDS: @sfount I've added this suggestion now. Let me know if you think it's alright and I'll merge!

PR #776, opened 2022-09-07
PP-10046: Minor docs fixes for September 2022

NathanD-GDS:
### Context
A collection of minor documentation changes that have accumulated over the past month or two.

This PR:

- makes API reference endpoint URLs consistent
- removes gaps between `status` values in the Search Agreements API reference
- specifies which `POST` requests don't need bodies
- changes the cancel payment error code descriptions
- removes a typo from the Get agreement endpoint in top-level API reference page
- adds new functionality to [what you can do with the API](https://docs.payments.service.gov.uk/quick_start_guide/#the-gov-uk-pay-api)
- removes the implication that Government Banking’s PSP could change soon from 'Take a payment'
- clarifies webhooks test account HTTP rules
- clarifies some wording around webhooks


SandorArpa (approved):
LGTM

PR #777, opened 2022-09-07
PP-10032: Updates happy-path diagram for API integration

NathanD-GDS:
### Context
We used a different tool to draw [the sequence diagram for recurring payments](https://docs.payments.service.gov.uk/recurring_payments/#understand-the-technical-flow-of-recurring-payments).

The new tool allowed us to make it look a bit more in-keeping with the rest of the docs and GDS style in general.

This means [that the other sequence diagram](https://docs.payments.service.gov.uk/integrate_with_govuk_pay/#when-to-release-your-service-to-your-users) is on our docs looks slightly dated and also doesn’t follow best practice for accessibility:

Updates the existing diagram to the new style.


SandorArpa (approved):
LGTM

PR #790, opened 2022-09-29
PP-9752 remove 3ds toggle option

SandorArpa:
### Context
- we no longer have an option to toggle 3ds on/off in selfservice. Remove any and all references.


NathanD-GDS (approved):
Looks good to me!

PR #816, opened 2022-12-08
RP docs changes based on Stroop feedback

NathanD-GDS:
### Context
As part of PP-10404, the Stroopwafel team had to read and feedback on the documentation for recurring payments and webhooks.

This PR:

- makes minor tweaks to the wording of recurring payments and webhooks documentation
- fixes `curl GET` to `curl -X GET` in various API reference pages


claireashworth (commented):
Pre-i complete. Looks good

source/recurring_payments/index.html.md.erb
```diff
@@ -57,7 +57,7 @@ In this integration example, the service creates an agreement when a user says t
 
 9. Your service [creates recurring payments using the `POST /v1/payments` endpoint](/recurring_payments/#taking-recurring-payments). In your request body, you  set `authorisation_mode` to `agreement` and include `"agreement_id": "{AGREEMENT_ID}"`, replacing `{AGREEMENT_ID}` with the agreement you set up earlier.
 
-10. GOV.UK Pay uses a token representing the user’s payment details to authorise the payment.
+10. GOV.UK Pay authorises the payment with the payment details the user entered for their first payment.
```
PippaClarkGDS, line 60:
```suggestion
10. GOV.UK Pay authorises the payment using the payment details the user entered for their first payment.
```

PippaClarkGDS: would 'using' work better here?

source/recurring_payments/index.html.md.erb
```diff
@@ -192,28 +193,27 @@ Your request should look something like this:
...
+  "reference": "CT-payment1",
   "authorisation_mode": "agreement",
   "agreement_id": "cgc1ocvh0pt9fqs0ma67r42l58"
 }
 ```
 
 The API returns a standard create payment response. Successful responses include the `agreement_id` you sent and `authorisation_mode` is `agreement`.
 
-GOV.UK Pay uses a token representing the user’s saved payment details from the agreement to automatically take the payment.
+GOV.UK Pay uses a token representing the user’s saved payment details from the agreement to take the payment.
```
PippaClarkGDS, line 204: is representing the right word here? Would reflecting or 'that captures' work better?

source/recurring_payments/index.html.md.erb
NathanD-GDS: I think it would, yes.

source/recurring_payments/index.html.md.erb
NathanD-GDS: Looking at other payment providers' docs, 'representing' seems fairly standard, as in [Stripe](https://stripe.com/docs/api/tokens?lang=go) and [Global Payments](https://www.globalpaymentsintegrated.com/en-us/blog/2020/05/05/payment-tokenization-guide).

There's a conversation within Pay to remove talk of 'tokenisation' anyway because services don't really need to know, so this line might be gone in a few days, anyway!

PippaClarkGDS (approved):

source/recurring_payments/index.html.md.erb
```diff
@@ -37,17 +37,17 @@ You can [read more about the lifecycle of agreements](/recurring_payments/#under
 
 In this integration example, the service creates an agreement when a user says they want to make recurring payments, then uses that agreement to continue taking payments.
 
-![A sequence diagram showing the journey of creating, setting up, and using an agreement to take recurring payments.](/images/recurring-payments-success-sequence.svg)
+![](/images/recurring-payments-success-sequence.svg)
```
kbottla, line 40: Is this change intentional? if so, any reason to not have Alt text here?

NathanD-GDS: This change is intentional. Because all of the information from the diagram is also in the body text, the diagram doesn't need alt text. If we include alt text, then we're duplicating information, which is bad for screen reader users.

I forgot why we didn't need it myself so I had to double check with the content community but you can see their answers here - https://gds.slack.com/archives/CACV3TACU/p1670937867761379

kbottla (approved):

PR #817, opened 2022-12-13
PP-10287 Update review date

kbottla:
## WHAT 
- Fact-checked and updated the review date for the following pages.
	- Delay taking a payment
	- Prefill payment fields
	- Connect your live account to Worldpay
	- Connect your live account to Stripe
	- Receive automatic payment event updates using webhooks
	- Take recurring payments
	- Add custom branding
- Any changes to these docs will be done separately.


SandorArpa (approved):
LGTM

PR #840, opened 2023-01-24
PP-10626: Add disclaimers to idempotency docs

NathanD-GDS:
### Context
Recurring payments supports idempotency to prevent services charging paying users multiple times for the same payment.

Idempotency is documented but we haven’t yet build our implementation of idempotency. Our implementation may change from what is currently documented.

This PR:

- removes some of the existing detail from the idempotency section of the recurring payments docs, in case our implementation changes
- adds a disclaimer to the top of [idempotency section](https://docs.payments.service.gov.uk/recurring_payments/#use-idempotency-to-avoid-double-charging-a-user), explaining that idempotency is not built yet and could change

### DIFF
--- a/source/recurring_payments/index.html.md.erb
+++ b/source/recurring_payments/index.html.md.erb
@@ -211,19 +211,24 @@ Read more about creating payments in [our guidance for taking payments](/making_
 
 ### Use idempotency to avoid double-charging a user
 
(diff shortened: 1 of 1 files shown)


stephencdaly (approved):
LGTM!

PR #844, opened 2023-01-30
BAU — Fix quotation marks on MOTO card details through API page

alexbishop1:
On the ‘Use an external system to take MOTO payments by sending card details through the API’ page, fix `"authorisation_mode": "moto_api"` so it has "straight quotation marks" rather than “curly quotation marks” (because it’s code and not a beautifully-typeset novel).


stephencdaly (approved):

NathanD-GDS:
@alexbishop1 This is an issue in a lot of places in the docs, so I'm going to replace all of the curly quotation marks this week.

PR #847, opened 2023-02-06
PP-10486 Update error codes for RCP

james-peacock-gds:
Context: Requests to create or setup an agreement, or take recurring payments, should only be possible from accounts with RCP enabled. Connector will return a RECURRING_CARD_PAYMENT_NOT_ALLOWED error if the
account does not have RCP enabled. This should map to a specific error (P0942) regardless of the endpoint called.

- Remove P2103 (which was created to cover just the 'create agreement' requests) and replace with P0942 (which covers create and setup agreements, and taking recurring payments).


SandorArpa (approved):
LGTM

PR #875, opened 2023-05-16
PP-10977 remove expired status from agreement

SandorArpa:
### Context

We don’t have any process/code that sets the payment_instrument/agreement status to expired. Also, we don’t have any plans to set payment instruments to expired status.


alexbishop1 (approved):
`expired` is expired

PR #882, opened 2023-05-18
BAU: Use npm ci instead of install, and make pr-build the same as build-and-deploy

jfharden:
### Context
build-and-deploy fails, build-pr does not fail.

Lets make them identical, and make it easier to diagnose the failed step by enabling -x in bash (to echo the command before running it)

1. Use npm ci instead of install
2. Use npm --omit=dev instead of --only=production (a warning tells you do to this when it runs --only=production)
3. Make build-pr perform the same steps as build-and-deploy for their build step.


nlsteers (approved):

SandorArpa (approved):
LOKTM

PR #886, opened 2023-05-18
BAU: Update to ruby 3.2

jfharden:
### Context
Ruby 2.7 is EOL. We were stuck from upgrading to 3 by middleman, but middleman now works with ruby 3, so lets update!


SandorArpa (approved):
LOKTM

jfharden:
🥳 Success!

PR #895, opened 2023-06-09
PP-11088: Adds RP test cards for Stripe accounts

NathanD-GDS:
### Context
We have test card numbers for Stripe test accounts for one-off payments in tech docs https://docs.payments.service.gov.uk/testing_govuk_pay/#if-you-39-re-using-a-test-stripe-account 

Stripe provides different card numbers to test recurring card payments which cover different scenarios https://stripe.com/docs/payments/save-during-payment?platform=web#web-test-the-integration and can be added to tech docs.

This PR:

- adds a new 'Testing recurring payments on a test Stripe account' heading
  - adds a table with mock cards for testing RPs on Stripe
- clarifies some wording for mock RP cards on GOV.UK Pay testing accounts
- explains earlier on the page that Stripe and gov.uk Pay test accounts have different sets of mock cards

### DIFF
--- a/source/testing_govuk_pay/index.html.md.erb
+++ b/source/testing_govuk_pay/index.html.md.erb
@@ -56,6 +56,11 @@ You can enter any valid value for other payment fields. For example, the card ex
 
 Mock card numbers do not work with your live account.
 
+There are separate sets of mock card numbers for:
+
+* [GOV.UK Pay test accounts](#if-you-39-re-using-a-test-39-sandbox-39-account) 
+* [Stripe test accounts](#if-you-39-re-using-a-test-stripe-account)
+
 #### If you're using a test ('sandbox') account
 
 |Testing action |Card number | Card brand | Card type |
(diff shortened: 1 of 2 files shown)


source/testing_govuk_pay/index.html.md.erb
```diff
@@ -80,15 +85,17 @@ Mock card numbers do not work with your live account.
 |Invalid CVC code|4000000000000127|Visa| Credit or debit |
 |General error|4000000000000119|Visa| Credit or debit |
 
-##### Testing recurring payments
+##### Testing recurring payments on a test account
```
kbottla, line 88: Looks like this change is breaking a link somewhere

kbottla (commented):
Changes look good to me. Just one small breaking change to be addressed.

source/testing_govuk_pay/index.html.md.erb
NathanD-GDS: @kbottla This should be fixed now

kbottla (approved):

PR #905, opened 2023-07-05
PP-11138: Configure Dependabot for security updates only

katstevens:
### Context
See RFC https://github.com/alphagov/pay-architecture/discussions/97

- Also adds config for Github Actions dependencies to make sure we pick up security updates for these.
- Also adds labels for npm, pip and bundler.


james-peacock-gds (approved):
LGTM

PR #926, opened 2023-10-18
PP-11669 - Clarifies where wallets are available and fees

NathanD-GDS:
### Context
Since enabling digital wallets for Stripe, we've had multiple support requests because users haven't been able to see the wallet buttons. We've also had multiple enquiries about potential additional fees for using Google Pay and/or Apple Pay.

This PR:

* clarifies the situations users can pay with Google Pay and Apple Pay
* clarifies that there are no additional fees for wallet payments

### DIFF
--- a/source/digital_wallets/index.html.md.erb
+++ b/source/digital_wallets/index.html.md.erb
@@ -7,14 +7,20 @@ weight: 2300
 
 # Take a digital wallet payment
 
-Your users can make payments with [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/).
+Your users can make payments with [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/). This page tells you how to enable digital wallet payments, test digital wallet payments, and the restrictions around diigtal wallets.
 
 Users can use digital wallets with:
 
 * payment links
 * payments created through the GOV.UK Pay API
 
+Users can only use Google Pay if they're making payments in Google Chrome.
+
+Users can only use Apple Pay if they're making payments in Safari on a recent Apple device.
+
 If your payment service provider (PSP) is Stripe, digital wallets are enabled by default. You can still disable Apple Pay and Google Pay in the GOV.UK Pay admin tool.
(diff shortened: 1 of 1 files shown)


source/digital_wallets/index.html.md.erb
```diff
@@ -7,15 +7,21 @@ weight: 2300
...
+Your users can make payments with [Google Pay](https://pay.google.com/intl/en_uk/about/) and [Apple Pay](https://www.apple.com/uk/apple-pay/). This page tells you how to enable digital wallet payments, test digital wallet payments, and the restrictions around diigtal wallets.
 
 Users can use digital wallets with:
 
 * payment links
 * payments created through the GOV.UK Pay API
 
+Users can currently only use Google Pay if they're making payments in Google Chrome.
+
+Users can only use Apple Pay if they're making payments in Safari on a recent Apple device.
```
alexbishop1, line 19: Apple Pay does already work in other browsers on iOS 16 (or at least it did in the betas) so “only … Safari” is already out of date: https://9to5mac.com/2022/08/01/apple-pay-web/

Maybe we could say something along the lines of this?

> Users can only use Google Pay if they’re making payments in a supported browser, such as Google Chrome.

> Users can only use Apple Pay if they’re making payments on a compatible Mac, iPhone or iPad in a supported browser, such as Safari.

Bit wordy but maybe you can tighten it a bit.

source/digital_wallets/index.html.md.erb
```diff
@@ -7,15 +7,21 @@ weight: 2300
...
+
 If your payment service provider (PSP) is Stripe, digital wallets are enabled by default. You can still disable Apple Pay and Google Pay in the GOV.UK Pay admin tool.
 
+There are no additional PSP fees if your service takes digital wallet payments.
```
alexbishop1, line 23: This is good.

alexbishop1 (changes requested):

source/digital_wallets/index.html.md.erb
NathanD-GDS: Wordy but I think it works. Pushed the changes, so feel free to re-review @alexbishop1

alexbishop1 (approved):

PR #928, opened 2023-10-31
BAU redirect the manual spaniel

nlsteers:
### Context

Self-evident


richblake (changes requested):
Hey Nat - Could you change it to #govuk-pay for now? We would like it to be everyone's problem, not just your team's.

richblake (approved):
Given recent discussion, approved.

PR #929, opened 2023-11-03
PP-11787: Minor docs changes based on feedback

NathanD-GDS:
### Context
We’ve identified a need for a few small documentation changes through reviews and support tickets. These changes are not large enough to need a whole ticket to themselves, so they’re aggregated into this one.

### The changes

- Corporate card fee page - remove inverted commas from response attribute names
- Use your own payment failure pages - link for payment status goes to the wrong page at the moment
- Use Welsh - link text currently implies that we have specific instructions for Welsh pages when the link should only go to the generic create payment docs page
- Testing integration - clarify the testing actions for recurring card payment cards
- Scheduled downtime - fix time formatting from 4am to 4:00am
- Go live with Stripe - make it clear that services need a real, physical address, and not a PO box
- Delayed capture - mention that it cannot be used with recurring payments
- Digital wallets - clarify that wallets respect any disabled card brands

### DIFF
--- a/source/corporate_card_surcharges/index.html.md.erb
+++ b/source/corporate_card_surcharges/index.html.md.erb
@@ -40,8 +40,8 @@ response fields:
 
 | Response field                   | Meaning
 |----------------------------------|-----------------------------------------------------------------------------------------------------------------------|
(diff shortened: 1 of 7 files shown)


SandorArpa (approved):
LGTM

PR #930, opened 2023-11-08
PP-11779: Removes mention of Stripe test accounts

NathanD-GDS:
### Context
Services can now test digital wallet payments in sandbox, so we no longer need to specify that services need a Stripe test account.

This PR:
- removes the mention of Stripe test accounts from the 'Take a digital wallet payment' docs page.


NathanD-GDS:
@stephencdaly Approve it now and then let me know when sandbox testing is available, then I'll merge.

Thanks!

stephencdaly (approved):

PR #934, opened 2023-11-14
PP-11813-deploy-tech-docs-to-github-pages

rfisher-gds:
### Context

We want to serve tech-docs from Github Pages rather than PAAS. 

Actions workflow to build and deploy static pages.

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,136 @@
+name: Static
+
+on:
+  push:
+    branches:
+      - 'master'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pages: write
+  id-token: write
+
+concurrency:
+  group: 'pages'
+  cancel-in-progress: true
+
+jobs:
+  run-tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/static.yml
```diff
@@ -0,0 +1,136 @@
...
+  group: 'pages'
+  cancel-in-progress: true
+
+jobs:
+  run-tests:
+    name: Unit tests and static build
+    uses: ./.github/workflows/run-tests.yml
+  static:
+    name:  Deploy and release Pay tech-doc pages
+    needs: run-tests
```
jfharden, line 24: We don't have a run-tests workflow in here 😬 

We should probably add one that just builds the site. I think you can find a run-tests workflow in the pay-product-pages repo too and just copy that into here also

jfharden (commented):
Just a small but important thing to address

jfharden (approved):

jfharden:
Approved knowing we still need to setup GitHub Pages and it's likely to fail on merge, but that we will enable it after then

PR #943, opened 2023-12-18
PP-11369: Adds information about webhook callback URL restrictions

NathanD-GDS:
### Context
Our documentation doesn't mention some of the restrictions we have on callback URLs for webhooks.

This PR clarifies that (from the JIRA ticket):

- test webhook callback URLs have no domain restrictions
- live webhook callback URLs must have a domain that ends in .gov.uk
- a service can request a domain be allowed for live webhook callback URLs by emailing support
- we will only allow domains that exclusively belong to the service or its parent organisation
- even we decide to allow a domain, it will not be added immediately


elenaschwan (commented):
Looks all good to me.

alexbishop1 (approved):
Looks good!

PR #948, opened 2024-01-29
PP-12126: Minor docs updates (January/February 2024)

NathanD-GDS:
### Context

Through docs reviews, support requests, and developer engagement, we’ve identified a number of small issues with our documentation. Each of these issues is not enough to merit its own Jira ticket, so they’re consolidated in this one.

* clarify where and when provider_id appears in API responses
* remove ‘conduct’ from ‘Integrate with the [GOV.UK](http://gov.uk/) Pay API’ for simplicity
* change **Enter card details** to **Enter payment details** to reflect new name of that page in the user journey


source/api_reference/single_payment_reference/index.html.md.erb
```diff
@@ -160,7 +160,7 @@ After you send your request to get information about a payment, you’ll receive
...
-| `provider_id`                                     | string          | The unique ID your payment service provider generated for this payment.<br><br>This is not the same as the `payment_id`.                                                                                                                                                                                                                         |
+| `provider_id`                                     | string          | The unique ID your payment service provider generated for this payment.<br><br>`provider_id` only appears after the user has submitted their payment details and GOV.UK Pay has tried to authorise the payment with your PSP.<br><br>This is not the same as the `payment_id`.                                                                                                                                                                                                                         |
```
ImogenCraigmile, line 163: I think it might be a very accepted term in the context of the pay docs but "authorise" to me doesn't sound like plain english, I prefer allow but in this context it might not be right so feel free too ignore me, just a thought!

source/integrate_with_govuk_pay/index.html.md.erb
```diff
@@ -43,8 +43,7 @@ You will likely need some kind of server-side datastore to record payment inform
 
 ## The GOV.UK Pay API
 
-The GOV.UK Pay API offers a set of operations to conduct and report on
-payments. For more information:
+The GOV.UK Pay API offers a set of endpoints to help your organisation reconcile and report on payments. For more information:
```
ImogenCraigmile, line 46: would integrate work instead of reconcile? Again feel free to ignore me just noodling words around in my head to think how they would sound!

ImogenCraigmile (approved):
I've left a couple of comments which are more just my thoughts and could be personal preference but in general the changes look really good and definitely work with our style guide!

source/integrate_with_govuk_pay/index.html.md.erb
NathanD-GDS: So reconciliation is a payment-specific thing. It's the process of matching the money going into your bank account to the payment entries in your database or whatever. It's not the clearest language though, completely agree. Don't think it can go anywhere here, though :'(

source/api_reference/single_payment_reference/index.html.md.erb
NathanD-GDS: Bit like reconciliation, 'authorise' is a specific (and not very plain English!) thing, so I don't think it can go. Curse these confusing words!

source/integrate_with_govuk_pay/index.html.md.erb
ImogenCraigmile: I thought this would be the case with both things but thought it worth raising! The user will understand and that is the main thing isn't it!

PR #959, opened 2024-03-26
PP-12078: Removed Postman and Insomnia recommendations

NathanD-GDS:
Change stems from docs review.

This PR removes Postman and Insomnia as recommended API clients for testing, because of recent security concerns.

Replaced with Bruno, the client the team has been widely using since Postman.


SandorArpa (approved):
LGTM

PR #961, opened 2024-04-03
Add new section with example emails services can send to their users

alanmaddrell:
Add new section with example emails services can send to their users

Has been checked by @NathanD-GDS


NathanD-GDS (approved):
Looks good to me.

PR #965, opened 2024-06-03
PP-12629 Update API docs with quoted field names

marcotranchino:
With this change, we are updating the public tech docs, making sure that all JSON fields are quoted.

Further information in JIRA.

https://payments-platform.atlassian.net/browse/PP-12629


alexbishop1 (approved):
`"Fantastic"`

PR #966, opened 2024-06-13
PP-12559 Migrate from google analytics to gtag

hjvoid:
### Context
Google is sunsetting Universal Analytics (GA3) at the end of June 2024.

The new version of analytics, GA4, recommends the use of Google Tag Manager. This is a functionality provided by Google, whereby arbitrary code may be entered into the Google Analytics console and then injected into a script on the application page.

The Google Tag Manager process bypasses Pay's [secure development requirements](https://manual.payments.service.gov.uk/manual/policies-and-procedures/software-development-lifecycle.html) (prioritisation, code written by trained software developers, code review and security testing) and is therefore not acceptable for use in any applications in the card data environment (CDE) or any applications that affect the CDE, as our PCI compliance would be at risk.

Google tag ('gtag') may be used as a static alternative to Google Tag Manager (see [Google's documentation for more information about the difference between the two](https://support.google.com/tagmanager/answer/7582054?hl=en)). This method does not include any code modifiable in the console and as such is suitable for use by GOV.UK Pay.

Implementation of gtag and removal of GA code.

### Guidance to review
See Jira ticket [PP-12559](https://payments-platform.atlassian.net/browse/PP-12559) . If you want to see this working locally you will need to rename the `application.min.js` file `application.js`

[PP-12559]: https://payments-platform.atlassian.net/browse/PP-12559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/source/javascripts/application.test.js
+++ b/source/javascripts/application.test.js
@@ -11,7 +11,7 @@ describe("Application JS", () => {
     document.cookie = "govuk_pay_cookie_policy=;expires=Thu, 01 Jan 1970 00:00:00 UTC;domain=.example.org";
 
     window.GovUkPay.cookies.showBannerIfConsentNotSet = jest.fn()
-    window.initialiseGoogleAnalytics = jest.fn()
+    window.initialiseGtag = jest.fn()
 
     // Override Document ready function in the test
     window.addEventListener = (event, callback) => {
@@ -33,7 +33,7 @@ describe("Application JS", () => {
     require("./application")
 
     expect(
-      window.initialiseGoogleAnalytics.mock.calls.length
+      window.initialiseGtag.mock.calls.length
       ).toBe(0);
   });
 
@@ -43,7 +43,7 @@ describe("Application JS", () => {
     require("./application")
 
     expect(
-      window.initialiseGoogleAnalytics.mock.calls.length
+      window.initialiseGtag.mock.calls.length
       ).toBe(0);
   });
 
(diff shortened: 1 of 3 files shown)


source/javascripts/application.test.js
```diff
@@ -22,39 +22,39 @@ describe("Application JS", () => {
   }); 
 
   it("should always call the showBannerIfConsentNotSet", () => {
-    require("./application")
+    require("./application.min.js")
```
iqbalgds, line 25: Just wondering if the original tests were still working?
As I was expecting this wouldn't need to be changed.

iqbalgds (changes requested):
Looks good - just got a minor question

source/javascripts/application.test.js
hjvoid: They didn't work for me  locally, but Github doesn't seem to mind whatever I put so I'm guessing they're not being run as part of the static build process?

iqbalgds: I would probably leave it as `/application`.
So this way, we are only getting your change in the PR.
I suspect that in github actions, it does not have the same issue.

iqbalgds (changes requested):
Just left a comment

source/javascripts/application.test.js
hjvoid: Cool. Will do!

iqbalgds (approved):
LGTM - good work

PR #979, opened 2024-10-23
Use DSv5 through Tech Docs Gem v.4.1.0

marcotranchino:
With this change, we are updating the Pay Tech Docs so that they use the new version of the Design System (v.5) through the brand new version of Tech Docs Gem[1].

Further information in Jira[2].

[1]
https://github.com/alphagov/tech-docs-gem/pull/375

[2]
https://payments-platform.atlassian.net/browse/PP-13122


SandorArpa (approved):
LOKTM

PR #985, opened 2024-12-18
PP-13369: Information about webhooks retry mechanism

NathanD-GDS:
### Context
Following the incident where webhook events were sent multiple times in error, we want to clarify to service users what the normal sending behaviour ('sent at least once') should be.

This PR adds some information to [the 'Retry mechanism' section of the GOV.UK Pay webhooks docs](https://docs.payments.service.gov.uk/webhooks/#retry-mechanism).


james-peacock-gds (approved):
LGTM

PR #995, opened 2025-02-21
BAU — Remove references to mountebank

alexbishop1:
Per https://github.com/bbyars/mountebank mountebank is no longer maintained and the mbtest.org domain is no longer controlled by the project so remove references to it (we still suggest WireMock).


richblake (approved):
Thanks Alex. LGTM!

PR #1021, opened 2025-09-22
Removes an irrelevant line about corporate fees

NathanD-GDS:
### Context
We do not, and do not plan to, support percentage-based corporate card fees.

This PR  removes a line about services asking about this


Beth-Brown (approved):
LGTM!

